PR #3, opened 2020-11-11
Fix undefined behaviour (shift into sign bit), and a bug, in varint functions

daviesrob:
First commit adds a test for the varint functions.  When I run them run on this commit, I get failures in `var_get_s32` and `var_get_s64`:

```
var_get_s32 failed:
Expected 32-bit 0x8f 0xff 0xff 0xff 0x7d => -2147483647
Got      32-bit 0x8f 0xff 0xff 0xff 0x7d => 1
var_get_s32 failed:
Expected 32-bit 0x8f 0xff 0xff 0xff 0x7e => 2147483647
Got      32-bit 0x8f 0xff 0xff 0xff 0x7e => -1
var_get_s32 failed:
Expected 32-bit 0x8f 0xff 0xff 0xff 0x7f => -2147483648
Got      32-bit 0x8f 0xff 0xff 0xff 0x7f => 0
var_get_s64 failed:
Expected 64-bit 0x81 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0x7d => -9223372036854775807
Got      64-bit 0x81 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0x7d => 1
var_get_s64 failed:
Expected 64-bit 0x81 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0x7e => 9223372036854775807
Got      64-bit 0x81 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0x7e => -1
var_get_s64 failed:
Expected 64-bit 0x81 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0x7f => -9223372036854775808
Got      64-bit 0x81 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0xff 0x7f => 0
```

The second commit fixes these functions, and also some illegal shift errors in the put functions reported by undefined behaviour sanitizer.

### DIFF
--- a/htscodecs/varint.h
+++ b/htscodecs/varint.h
@@ -267,20 +267,26 @@ static inline int var_get_u32(uint8_t *cp, const uint8_t *endp, uint32_t *i) {
 // This folds the sign bit into the bottom bit so we iterate
 // 0, -1, +1, -2, +2, etc.
 static inline int var_put_s32(uint8_t *cp, const uint8_t *endp, int32_t i) {
-    return var_put_u32(cp, endp, (i << 1) ^ (i >> 31));
+    uint32_t u = (uint32_t) i;
+    return var_put_u32(cp, endp, (u << 1) ^ (i >= 0 ? 0U : ~0U));
 }
 static inline int var_put_s64(uint8_t *cp, const uint8_t *endp, int64_t i) {
-    return var_put_u64(cp, endp, (i << 1) ^ (i >> 63));
+    uint64_t u = (uint64_t) i;
+    return var_put_u64(cp, endp, (u << 1) ^ (i >= 0 ? 0ULL : ~0ULL));
 }
 
 static inline int var_get_s32(uint8_t *cp, const uint8_t *endp, int32_t *i) {
-    int b = var_get_u32(cp, endp, (uint32_t *)i);
-    *i = (*i >> 1) ^ -(*i & 1);
+    uint32_t u;
+    int b = var_get_u32(cp, endp, &u);
+    int32_t iv = (int32_t)(u >> 1);
+    *i = (u & 1) ? (-iv - 1) : iv;
     return b;
 }
 static inline int var_get_s64(uint8_t *cp, const uint8_t *endp, int64_t *i) {
-    int b = var_get_u64(cp, endp, (uint64_t *)i);
-    *i = (*i >> 1) ^ -(*i & 1);
+    uint64_t u;
+    int b = var_get_u64(cp, endp, &u);
+    int64_t iv = (int64_t)(u >> 1);
+    *i = (u & 1) ? (-iv - 1) : iv;
     return b;
 }
 
@@ -295,7 +301,8 @@ static inline int var_size_u64(uint64_t v) {
 #define var_size_u32 var_size_u64
 
 static inline int var_size_s64(int64_t v) {
-    return var_size_u64((v >> 63) ^ (v << 1));
+    uint64_t u = (uint64_t) v;
+    return var_size_u64((u << 1) ^ (v >= 0 ? 0ULL : ~0ULL));
 }
 #define var_size_s32 var_size_s64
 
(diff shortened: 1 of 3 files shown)


jkbonfield:
The test is useful, but the fix unfortunately trades efficiency for pedantry.  I get the point that yes this is undefined behaviour (although I couldn't get gcc to label it as such, try hard as I may), but only in as far as assuming 2s complement is undefined behaviour.  Correct use of a cast changes this to implementation defined behaviour, which is a far better solution.

The techniques are widely used by many more important things than htslib.  Eg maybe you'd like to file a bug against [Google Protobuf too](https://github.com/protocolbuffers/protobuf/blob/master/ruby/ext/google/protobuf_c/upb.h#L6273-L6285)? (Also see https://developers.google.com/protocol-buffers/docs/encoding)

I'm not willing to sacrifice performance in order for a warm fuzzy feeling.  Your revised get_s32 code introduces branching on icc, MSVC and the PowerPC compiler) and a conditional move instruction on clang / gcc).

There is however a bug which I thank you for spotting.   It's caused by the incorrect use of signed vs unsigned leading to a "sar" instead of "shr" instruction.  The fixes I think ideally should be:

```
static inline int var_put_s32(uint8_t *cp, const uint8_t *endp, int32_t i) {
    return var_put_u32(cp, endp, ((uint32_t)i << 1) ^ (i >> 31));
} 

static inline int var_get_s32(uint8_t *cp, const uint8_t *endp, int32_t *i) {
    int b = var_get_u32(cp, endp, (uint32_t *)i);
    *i = ((uint32_t)*i >> 1) ^ -(int32_t)(*i & 1);
    return b;
}
```

(And similar 64-bit)

These migrate it from undefined behaviour to implementation defined behaviour, which I'm happy with given it's basically just an assumption on the machine using 2s-complement arithmetic.

jkbonfield:
Merged 1st commit and rewrote second as 44ec7a9

daviesrob:
Yes, the protobuf code probably does need a bug filing against it.  The problem is very much in the part that does `n >> 31` and expects that the result will be an arithmetic shift and not a logical one.  There is absolutely no guarantee that will happen - it completely depends on an arbitrary choice made by the compiler writer.

If you really want it to work that way, you would need to write the code in assembler so that you know exactly which instruction will be used.

daviesrob:
Also see https://godbolt.org/z/Gvc3s1

The two versions compile, with optimisation, to near identical assembly.  So there is absolutely no penalty for having pedantically-correct code, and much benefit from knowing that it won't suddenly stop working in the future.

jkbonfield:
I've already been to godbolt and done my own tests: https://godbolt.org/z/5s5TPG

As you can see, it does introduce conditional branches on several compilers.  I did my research before deciding which method to use.

daviesrob:
Yes, the icc version of the get function is a bit disappointing, and the replacement for that isn't so bad as it only relies on implementation-defined behaviour.  It could be implemented with rather less casting like this, though:
```
int var_get_s32b(uint8_t *cp, const uint8_t *endp, int32_t *i) {
    uint32_t u;
    int b = var_get_u32(cp, endp, &u);
    *i = (int32_t)((u >> 1) ^ -(u & 1));
    return b;
}
```

The put function however relies on undefined behaviour, which is rather more serious.  As all the compilers seem to produce the same, or very similar, code for the safe version I don't see why you wouldn't want to use that one.

jkbonfield:
Which also produces poorer code (I checked that too before).  It may be close enough it doesn't matter though, but the optimal version is shorter so I see no reason to change it again.

The put isn't undefined.  It's implementation defined.  It has a << on unsigned which is 100% defined.  It has "i >> 31" which is implementation defined.  Like I say, I have no desire to fix things to work on non 2s-complement machines.  If we ever get one of those we have far bigger problems than tinkering with htscodecs!

PR #13, opened 2021-01-25
Removed undefined shift (malformed input data).

jkbonfield:
Credit to OSS-Fuzz
Fixes oss-fuzz 29817

### DIFF
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -611,7 +611,7 @@ unsigned char *rans_uncompress_O0_4x16(unsigned char *in, unsigned int in_size,
     // Build symbols; fixme, do as part of decode, see the _d variant
     for (j = x = 0; j < 256; j++) {
 	if (F[j]) {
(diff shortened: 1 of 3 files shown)


htscodecs/rANS_static4x16pr.c
```diff
@@ -1118,7 +1118,7 @@ unsigned char *rans_uncompress_O1_4x16(unsigned char *in, unsigned int in_size,
 	// Build symbols; fixme, do as part of decode, see the _d variant
 	for (j = x = 0; j < 256; j++) {
 	    if (F[j]) {
-		if (x + F[j] > (1<<shift))
+		if (F[j] > (1<<shift) - x)
```
daviesrob, line 1121: Could a similar problem to this also occur [here](https://github.com/samtools/htscodecs/blob/e8d8597e7/htscodecs/rANS_static4x16pr.c#L614)?

jkbonfield: Yes.  I'll amend that too.

PR #19, opened 2021-02-09
Use printf in shell scripts instead of echo -n

jmarshall:
On macOS the shell's builtin `echo` doesn't process options in some circumstances, resulting in

```
-n Testing arith_dynamic [...]
11615
```

as can be seen in [HTSlib macOS CI build logs](https://cirrus-ci.com/task/4839257728614400).

Using `printf(1)` is (sadly) overall more portable.

### DIFF
--- a/tests/arith.test
+++ b/tests/arith.test
@@ -15,7 +15,7 @@ do
 	then
 	    continue
 	fi
-        echo -n Testing arith_dynamic -r -o$o on $f "	"
+        printf 'Testing arith_dynamic -r -o%s on %s\t' $o "$f"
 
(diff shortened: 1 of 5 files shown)


jkbonfield:
Indeed what a sad state of affairs, but even POSIX admit echo isn't portable and recommend printf.  https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html#tag_20_37_16

Thanks for the fix.

jmarshall:
As a historical application myself, I'll of course carry on using `echo` for the usual case when I want a newline at the end :smile:

PR #21, opened 2021-02-23
Release 1.0

jkbonfield:
Preparation for release 1.0.  We can edit the date on merge day if deemed necessary.

Since 0.5 no new structures or public functions have arrived / changed, so this is just a revision update as far as ABI goes.

There is a new macro for testing version: HTSCODECS_VERSION.  That's in a new <htscodecs/htscodecs.h> file.  Minimal right now, but it has to go somewhere.

Also bundled some other minor and largely irrelevant tidyups.


htscodecs/htscodecs.h
```diff
@@ -0,0 +1,44 @@
...
+#define HTSCODECS_H
+
+/*
+ * Version X.Y.Z encoded as XYYYZZZD
+ * where D is 0 for official releases and *may* be >0 for
+ * untagged github checkouts.
+ */
+#define HTSCODECS_VERSION 10000000
+
+#endif */ HTSCODECS_H */
```
jmarshall, line 44: Borked comment. Does nothing in htscodecs include _htscodecs.h_ and/or is there an opportunity to add a suitable `‑Werror` build to the CI?

jkbonfield: Good spot!  Thanks

Werror wouldn't have any change as nothing actually includes this yet.  Maybe I should do that somewhere just to check it. :)

I'm also thinking maybe the X.Y.Z isn't needed as I don't really plan on using the .Z, so can use the htslib style for that perhaps.

htscodecs/htscodecs.h
```diff
@@ -0,0 +1,45 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef HTSCODECS_H
+#define HTSCODECS_H
+
+/*
+ * Version X.Y.Z encoded as XYYYZZZ.
```
jmarshall, line 38: The HTSlib style follows Boost in having only two digits of ZZ, so XYYYZZ here would mean it was encoded exactly the same way. (Also “in between” is two words, or use just “between”.)

jkbonfield: Thanks for the comments John.  In the interests of consistency, I'll change it to match htslib's pattern then.  It's highly unlikely we'll be using more than 100 patch-levels.

I'm not sure yet if I want to bother with the whole "90" bit for unreleased github changes yet, as I'm not expecting it to be changing often once it's settled down.

jmarshall: Cool, thanks. :+1: re the whole "90" bit; I imagine it'll settle down more too.

daviesrob:
An `htscodecs_version()` function would be a useful partner to the `HTSCODECS_VERSION` macro.  The run-time version may not be the one you had at compile time - especially if htscodecs is linked as a DLL.

jkbonfield:
As requested, I added an `htscodecs_version` function.  It's a little bit complex as I'm assuming this is a request to do the same style as htslib, which means with and without git, autobuilding the config, etc.  It also means an additional file (htscodecs.c) because until now it's been nothing more than a collection of codecs so there's no general purpose package source, unless we pick on utils.h, which wouldn't fit the requirement to return the version of the dynamic library found rather than one compiled against.

So now we can distinguish run-time object (func) vs compiler-time (macro).  There's some nuance in that htscodecs.h perhaps should also be generated on-the-fly too from a .in file, to limit updating version to one single source file.  For now it's manual though.

htscodecs/utils.h
```diff
@@ -32,6 +32,7 @@
  */
 
 #include <string.h>
+#include "htscodecs.h" // to check for syntax errors only
```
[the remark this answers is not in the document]
jkbonfield: Good point.  I'll correct that and the .PHONY bit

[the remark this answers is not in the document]
jkbonfield: Both now done, although I don't understand what adding to .PHONY actually changes here as it worked without it.  (It fired when required, and didn't when not.)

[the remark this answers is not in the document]
daviesrob: It protects you from someone making a file called `force`.

PR #22, opened 2021-03-09
Fix spurious test failure on Solaris/OpenIndiana

jmarshall:
This fails with `awk: record 'S9$#6#6%7654A5634A;$...' too long` on Solaris/OpenIndiana's `awk`. We could configure around this to use `nawk` on this platform instead (as in PR samtools/samtools#1165, but more complicated due to needing to transmit the setting from htslib as well), but in this case it's easier to just use `cut`.

### DIFF
--- a/tests/fqzcomp.test
+++ b/tests/fqzcomp.test
@@ -8,7 +8,7 @@ fi
 for f in `ls -1 $srcdir/dat/q* $srcdir/htscodecs-corpus/dat/q* 2>/dev/null`
 do
     comp=${f%/*/*}/dat/fqzcomp/${f##*/}
-    awk '{print $1}' < $f > $out/fqz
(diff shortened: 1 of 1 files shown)


jkbonfield:
Wow that's really quite a poor showing for Solaris!  It never ceases to amaze me how these distributions cling to brain dead implementations of tools when there are better quality free versions out there.  The longest line is under 14Kb, but apparently [Solaris can't cope with more than 2559 bytes](https://balazsdeak.blogspot.com/2010/02/solaris-awk-max-record-size-is-2559-2-8.html)!

However yes, this test is just showing my penchant for using awk as my go-to processing tool.  Cut would be prefereable indeed. :-)

jmarshall:
I didn't even look at how long that particular line was, but that is crappy indeed! :smile:

PR #24, opened 2021-05-14
Fix a bug with name tokeniser and variable number of names per slice.

jkbonfield:
If the number of names increases by exactly 1 between successive slices, and the old value was the size of our allocated buffer, then we don't grow it by one.  This causes the last decoded name to return -1 instead of 0 as we do the max_names check before doing t0 < 0, which in turn gives a failure to decode slice error message.

It could be fixed in two ways, by shifting where the end of names check happens, or fixing the allocation.  I chose the latter as it would take more head scratching to determine why the allocation has `++max_names` in it.

### DIFF
--- a/htscodecs/tokenise_name3.c
+++ b/htscodecs/tokenise_name3.c
@@ -211,7 +211,7 @@ static name_context *create_context(int max_names) {
 	if (!ctx) return NULL;
 	ctx->max_names = max_names;
 	pthread_setspecific(tok_key, ctx);
-    } else if (ctx->max_names < max_names) {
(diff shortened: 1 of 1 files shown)


jkbonfield:
Sanger local users can see a test file in ~jkb/scratch/data/broken_in_v31.sam.  It puts 405 seqs in the first slice and 406 in the second.  (The number is due to the bases-per-slice limit being hit, well before the seqs-per-slice one, and this is the cause of the variable number of records used.)

daviesrob:
Looks OK, although your test cram file worked on both master and this PR for me.

PR #25, opened 2021-06-01
Modernise autoconf usage

jmarshall:
Avoid warnings about obsolete usage from `autoreconf`:

```
configure.ac:73: warning: The macro `AC_PROG_LIBTOOL' is obsolete.
configure.ac:73: You should run autoupdate.
configure.ac:93: warning: The macro `AC_HELP_STRING' is obsolete.
configure.ac:93: warning: The macro `AC_LANG_C' is obsolete.
configure.ac:111: warning: The macro `AC_HEADER_STDC' is obsolete.
configure.ac:119: warning: AC_OUTPUT should be used without arguments.
```

The most recent “new” autoconf feature in this change is from 2.59 in 2006. `LT_INIT` has sufficed for libtool for even longer.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -70,7 +70,6 @@ dnl Checks for programs.
 AC_PROG_CC
 AC_PROG_INSTALL
 LT_INIT
-AC_PROG_LIBTOOL
 
 # # by Marcelo Magallon <mmagallo@efis.ucr.ac.cr>
 # # Turn around -rpath problem with libtool 1.0c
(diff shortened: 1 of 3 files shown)


jkbonfield:
Sadly our systems have 2.69 as the latest version!  However it works on the modified files so thanks for improving this.

jmarshall:
This was with 2.71, which has been what you get with Brew on macOS for a while. I didn't know how new those warnings were, but having now tried it in 2.69 I see the answer is: all of them are new in 2.70+!

But yes, happily the “new” macros recommended by the warnings have been around a lot longer than that.

PR #26, opened 2021-06-17
Experimental 32-way unrolled 16-bit rANS codec

jkbonfield:
The extra level of unrolling permits good use of SIMD instructions, particularly favouring AVX2 and AVX512, but also showing some benefit from ARM Neon.  It is likely WebAssembly may have some benefits too given this is also a 128-bit vector (as Neon).

None of these are capable of being auto-vectorised, not by gcc, clang nor icc.  Hence all implementations are hand written.  I have benchmarks for 4-way, 32-way scalar, and 32-way vectorised with various methods, posted below.

Questions:

- Given the 32-way scalar code is slower than the 4-way scalar code, and the compilers cannot auto-vectorise, how much emphasis do we put on maximum performance in C compared to potential slow downs in other languages (given this is a file-format visible change).

- Do we wish to do this for order-0 only, or also order-1?  Order-1 is more questionable looking at the benchmarks.  It definitely gains on Intel, but minimally on ARM.


jkbonfield:
Benchmarks on multiple systems:

Input is ~/scratch/data/q40

--------------------
Farm5-gpu0101:  Intel(R) Xeon(R) Gold 6142 CPU @ 2.60GHz

    4x8 -o0        325.6 MB/s enc, 477.2 MB/s dec    101000000 bytes -> 51140911 bytes
    4x16 -o0       412.0 MB/s enc, 776.1 MB/s dec    101000000 bytes -> 51136478 bytes
    32x16 -o4      255.4 MB/s enc, 622.0 MB/s dec    101000000 bytes -> 51144602 bytes
     " avx2        506.6 MB/s enc, 1756.1 MB/s dec   101000000 bytes -> 51144602 bytes
     " avx512      751.8 MB/s enc, 2099.6 MB/s dec   101000000 bytes -> 51144602 bytes
    
    4x8 -o1        225.8 MB/s enc, 328.7 MB/s dec    101000000 bytes -> 48944198 bytes
    4x16 -o1/10    222.9 MB/s enc, 471.3 MB/s dec    101000000 bytes -> 48928863 bytes
    32x16 -o5/10   200.3 MB/s enc, 386.1 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx2        239.7 MB/s enc, 951.3 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx512      346.9 MB/s enc, 1071.1 MB/s dec   101000000 bytes -> 48937630 bytes
    
    4x16 -o1/12    224.7 MB/s enc, 387.7 MB/s dec    101000000 bytes -> 48874315 bytes
    32x16 -o5/12   200.8 MB/s enc, 321.9 MB/s dec    101000000 bytes -> 48883012 bytes
     " avx2        237.2 MB/s enc, 644.7 MB/s dec    101000000 bytes -> 48883012 bytes
     " avx512      341.7 MB/s enc, 712.1 MB/s dec    101000000 bytes -> 48883012 bytes

--------------------
node-10-1-1: Intel(R) Xeon(R) Gold 6242 CPU @ 2.80GHz

    4x8 -o0        347.1 MB/s enc, 514.3 MB/s dec    101000000 bytes -> 51140911 bytes
    4x16 -o0       436.1 MB/s enc, 819.4 MB/s dec    101000000 bytes -> 51136478 bytes
    32x16 -o4      269.2 MB/s enc, 654.6 MB/s dec    101000000 bytes -> 51144602 bytes
     " avx2        528.6 MB/s enc, 1851.7 MB/s dec   101000000 bytes -> 51144602 bytes
     " avx512      785.6 MB/s enc, 2205.5 MB/s dec   101000000 bytes -> 51144602 bytes
    
    4x8 -o1        239.7 MB/s enc, 348.7 MB/s dec    101000000 bytes -> 48944198 bytes
    4x16 -o1/10    233.2 MB/s enc, 493.4 MB/s dec    101000000 bytes -> 48928863 bytes
    32x16 -o5/10   209.1 MB/s enc, 415.6 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx2        244.1 MB/s enc, 1001.7 MB/s dec   101000000 bytes -> 48937630 bytes
     " avx512      361.5 MB/s enc, 1127.7 MB/s dec   101000000 bytes -> 48937630 bytes
    
    4x16 -o1/12    232.5 MB/s enc, 401.0 MB/s dec    101000000 bytes -> 48874315 bytes
    32x16 -o5/12   207.0 MB/s enc, 338.5 MB/s dec    101000000 bytes -> 48883012 bytes
     " avx2        244.6 MB/s enc, 634.5 MB/s dec    101000000 bytes -> 48883012 bytes
     " avx512      360.8 MB/s enc, 680.7 MB/s dec    101000000 bytes -> 48883012 bytes

--------------------
bc-27-2-06:  AMD Opteron(TM) Processor 6272 (no avx2/avx512)

    4x8 -o0        140.1 MB/s enc, 163.5 MB/s dec    101000000 bytes -> 51140911 bytes
    4x16 -o0       186.1 MB/s enc, 205.8 MB/s dec    101000000 bytes -> 51136478 bytes
    32x16 -o4      118.4 MB/s enc, 203.0 MB/s dec    101000000 bytes -> 51144602 bytes
     " avx2        (Not available)
    
    4x8 -o1         67.1 MB/s enc,  94.2 MB/s dec    101000000 bytes -> 48944198 bytes
    4x16 -o1/10     66.3 MB/s enc, 133.0 MB/s dec    101000000 bytes -> 48928863 bytes
    32x16 -o5/10    56.6 MB/s enc,  84.4 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx2        (Not available)
    
    4x16 -o1/12     68.7 MB/s enc, 101.5 MB/s dec    101000000 bytes -> 48874315 bytes
    32x16 -o5/12    55.0 MB/s enc,  65.7 MB/s dec    101000000 bytes -> 48883012 bytes

--------------------
deskpro107386: Intel(R) Core(TM) i5-4570 CPU @ 3.20GHz

    4x8 -o0        307.6 MB/s enc, 549.5 MB/s dec    101000000 bytes -> 51140911 bytes
    4x16 -o0       390.4 MB/s enc, 657.1 MB/s dec    101000000 bytes -> 51136478 bytes
    32x16 -o4      251.0 MB/s enc, 543.2 MB/s dec    101000000 bytes -> 51144602 bytes
     " avx2        501.1 MB/s enc, 1268.5 MB/s dec   101000000 bytes -> 51144602 bytes
     " avx512      (Not available)
    
    4x8 -o1        209.2 MB/s enc, 309.9 MB/s dec    101000000 bytes -> 48944198 bytes
    4x16 -o1/10    203.4 MB/s enc, 443.1 MB/s dec    101000000 bytes -> 48928863 bytes
    32x16 -o5/10   189.2 MB/s enc, 351.4 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx2        220.5 MB/s enc, 808.0 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx512      (Not available)
    
    4x16 -o1/12    203.0 MB/s enc, 344.4 MB/s dec    101000000 bytes -> 48874315 bytes
    32x16 -o5/12   189.2 MB/s enc, 277.0 MB/s dec    101000000 bytes -> 48883012 bytes
     " avx2        219.8 MB/s enc, 550.3 MB/s dec    101000000 bytes -> 48883012 bytes

--------------------
Home PC: Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz (via WSL)

    4x8 -o0        303.4 MB/s enc, 492.7 MB/s dec    101000000 bytes -> 51140911 bytes
    4x16 -o0       461.5 MB/s enc, 780.2 MB/s dec    101000000 bytes -> 51136478 bytes
    32x16 -o4      298.6 MB/s enc, 656.3 MB/s dec    101000000 bytes -> 51144602 bytes
     " avx2        590.7 MB/s enc, 1524.6 MB/s dec   101000000 bytes -> 51144602 bytes
     " avx512      (Not available)
    
    4x8 -o1        217.2 MB/s enc, 295.0 MB/s dec    101000000 bytes -> 48944198 bytes
    4x16 -o1/10    238.1 MB/s enc, 517.0 MB/s dec    101000000 bytes -> 48928863 bytes
    32x16 -o5/10   221.1 MB/s enc, 419.1 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx2        260.7 MB/s enc, 929.5 MB/s dec    101000000 bytes -> 48937630 bytes
     " avx512      (Not available)
    
    4x16 -o1/12    237.0 MB/s enc, 401.2 MB/s dec    101000000 bytes -> 48874315 bytes
    32x16 -o5/12   222.9 MB/s enc, 329.7 MB/s dec    101000000 bytes -> 48883012 bytes
     " avx2        259.2 MB/s enc, 629.6 MB/s dec    101000000 bytes -> 48883012 bytes

--------------------
ARM: AWS Graviton-2 (A72 based)

    4x8 -o0        373.0 MB/s enc, 480.1 MB/s dec    152000000 bytes -> 10989766 bytes
    4x16 -o0       347.6 MB/s enc, 452.6 MB/s dec    101000000 bytes -> 51136478 bytes
    32x16 -o4      343.8 MB/s enc, 437.3 MB/s dec    101000000 bytes -> 51144602 bytes
     " neon        365.4 MB/s enc, 736.2 MB/s dec    101000000 bytes -> 51144602 bytes
    
    4x8 -o1        ?
    4x16 -o1/10    205.1 MB/s enc, 333.4 MB/s dec    101000000 bytes -> 48928863 bytes
    32x16 -o5/10   168.0 MB/s enc, 263.1 MB/s dec    101000000 bytes -> 48937630 bytes
     " neon        175.5 MB/s enc, 386.3 MB/s dec    101000000 bytes -> 48937630 bytes
    
    4x16 -o1/12    205.5 MB/s enc, 286.6 MB/s dec    101000000 bytes -> 48874315 bytes
    32x16 -o5/12   162.6 MB/s enc, 216.9 MB/s dec    101000000 bytes -> 48883012 bytes
     " neon        177.5 MB/s enc, 325.3 MB/s dec    101000000 bytes -> 48937630 bytes

(Double check: scalar size differs to neon)

jkbonfield:
Compiler bake off:

Compiler differences on Intel(R) Xeon(R) Gold 6242 CPU @ 2.80GHz

4x16 -o0:

    gcc-7     431.5 MB/s enc, 820.1 MB/s dec         101000000 bytes -> 51136478 bytes
    gcc-11    268.2 MB/s enc, 718.3 MB/s dec         101000000 bytes -> 51136478 bytes
    clang-10  393.7 MB/s enc, 827.3 MB/s dec         101000000 bytes -> 51136478 bytes
    icc-19    383.6 MB/s enc, 556.8 MB/s dec         101000000 bytes -> 51136478 bytes

gcc is bad at encode speed, icc is bad at decode speed, clang is best at both.
gcc-7 is Ubuntu 18 default (7.5.0).  Far better than gcc-11.

This is the only test I'd say a compiler is particularly bad at compared to the others.
(Exception: clang on ARM)

TODO: research why gcc-11 is poor at o0 encode compared to gcc-7.
Can we adjust the code somehow?

32x16 -o0x605 (O0 scalar)

    gcc-7     208.9 MB/s enc, 413.5 MB/s dec         101000000 bytes -> 48937630 bytes
    gcc-11    206.5 MB/s enc, 422.4 MB/s dec         101000000 bytes -> 48937630 bytes
    clang-10  203.0 MB/s enc, 392.2 MB/s dec         101000000 bytes -> 48937630 bytes
    icc-19    203.5 MB/s enc, 379.5 MB/s dec         101000000 bytes -> 48937630 bytes

None of them bad now, but gcc-11 slightly ahead.

32x16 -o0x805 (O0 AVX2)

    gcc-7     245.2 MB/s enc, 1006.2 MB/s dec        101000000 bytes -> 48937630 bytes
    gcc-11    241.3 MB/s enc, 1004.2 MB/s dec        101000000 bytes -> 48937630 bytes
    clang-10  275.9 MB/s enc, 999.2 MB/s dec         101000000 bytes -> 48937630 bytes
    icc-19    258.2 MB/s enc, 973.2 MB/s dec         101000000 bytes -> 48937630 bytes

Again none are bad, but clang edging ahead.

32x16 -o0x5 (O0 AVX512)

    gcc-7     375.1 MB/s enc, 1135.4 MB/s dec        101000000 bytes -> 48937630 bytes
    gcc-11    371.4 MB/s enc, 1189.2 MB/s dec        101000000 bytes -> 48937630 bytes
    clang-10  400.2 MB/s enc, 1204.7 MB/s dec        101000000 bytes -> 48937630 bytes
    icc-19    346.2 MB/s enc, 1175.3 MB/s dec        101000000 bytes -> 48937630 bytes

All very close for decode and none bad encode, clang best.

4x16 -o1:

    gcc-7     228.7 MB/s enc, 495.1 MB/s dec         101000000 bytes -> 48928863 bytes
    gcc-11    240.8 MB/s enc, 512.3 MB/s dec         101000000 bytes -> 48928863 bytes
    clang-10  238.6 MB/s enc, 512.8 MB/s dec         101000000 bytes -> 48928863 bytes
    icc-19    234.9 MB/s enc, 450.8 MB/s dec         101000000 bytes -> 48928863 bytes

gcc/clang very close, with icc dropping a bit on decode.
gcc-7 is slightly behind gcc-11 here.

32x16 -o0x605 (O1 scalar)

    gcc-7     208.3 MB/s enc, 406.5 MB/s dec         101000000 bytes -> 48937630 bytes
    gcc-11    204.5 MB/s enc, 405.2 MB/s dec         101000000 bytes -> 48937630 bytes
    clang-10  203.4 MB/s enc, 398.4 MB/s dec         101000000 bytes -> 48937630 bytes
    icc-19    207.6 MB/s enc, 385.1 MB/s dec         101000000 bytes -> 48937630 bytes

All comparable.

32x16 -o0x805 (O1 AVX2)

    gcc-7     244.7 MB/s enc, 1007.5 MB/s dec        101000000 bytes -> 48937630 bytes
    gcc-11    245.9 MB/s enc, 1009.2 MB/s dec        101000000 bytes -> 48937630 bytes
    clang-10  282.1 MB/s enc, 1001.3 MB/s dec        101000000 bytes -> 48937630 bytes
    icc-19    255.7 MB/s enc, 969.9 MB/s dec         101000000 bytes -> 48937630 bytes

Clang edging ahead on encode speed.

32x16 -o0x5 (O1 AVX512)

    gcc-7     371.3 MB/s enc, 1128.5 MB/s dec        101000000 bytes -> 48937630 bytes
    gcc-11    371.4 MB/s enc, 1189.3 MB/s dec        101000000 bytes -> 48937630 bytes
    clang-10  398.5 MB/s enc, 1206.4 MB/s dec        101000000 bytes -> 48937630 bytes
    icc-19    345.0 MB/s enc, 1176.5 MB/s dec        101000000 bytes -> 48937630 bytes

Clang edging ahead on encode speed, where icc lags a bit.



For ARM, gcc vs clang-10 is noticably different when Neon instructions
are enabled:

Order-0 Neon

    gcc     365.4 MB/s enc, 736.2 MB/s dec   101000000 bytes -> 51144602
    clang   317.2 MB/s enc, 528.5 MB/s dec   101000000 bytes -> 51144602

Order-1 Neon

    gcc     175.5 MB/s enc, 386.3 MB/s dec   101000000 bytes -> 48937630
    clang   179.2 MB/s enc, 347.0 MB/s dec   101000000 bytes -> 48937630

This is a particular poor showing for order-0 clang.
The non-Neon versions are almost neck and neck.  The difference is
Clang is reordering the Neon intrinsics and apparently introducing
either instruction latency or port contention.

jkbonfield:
As a final "what yet more?" question, it may be worth considering if rANS is even the right approach for large scale SIMD vectorisation.  Possibly tANS would scale better.  I've not explored this yet.

jkbonfield:
Major squashing and rebased **on top of the speedups PR** (which is the reason for squashing, as those basic non-SIMD oriented speed ups have been pulled out of this PR so they can arrive earlier if desired).

Note the old copy of unroll32 branch before all the hacking around is available at https://github.com/jkbonfield/htscodecs/tree/unroll32-old

htscodecs/rANS_static16_int.h
```diff
@@ -0,0 +1,633 @@
...
+static inline int decode_freq(uint8_t *cp, uint8_t *cp_end, uint32_t *F,
+			      uint32_t *fsum) {
+    if (cp == cp_end)
+	return 0;
+
+    uint8_t *op = cp;
+    cp += decode_alphabet(cp, cp_end, F);
+
+    int j, tot = 0;
+    for (j = 0; j < 256; j++) {
```
daviesrob, line 262: This doesn't quite follow the specification, as it assumes an ordering on the list returned by the `read_alphabet` spec. function.  While the way the alphabet run-length encoding works implies that symbols will be increasing order, the spec. doesn't appear to prevent you from storing them using some other ordering.  If that did happen, this implementation would assign the wrong frequency to each symbol.

Either the specification should be adjusted to say that the alphabet is always stored in increasing order (or interpreted as if it was); or this code should allow for different orderings.  The former may be the best option, as I suspect a non-increasing order isn't that useful.

daviesrob: [Related hts-specs comment](https://github.com/samtools/hts-specs/pull/433/files#r814844673)

htscodecs/rANS_static4x16pr.c
```diff
@@ -100,282 +89,25 @@
...
-	T += f;
-    }
-
-    if (total) *total = T;
-    return cp - op;
-}
-
-unsigned int rans_compress_bound_4x16(unsigned int size, int order) {
-    int N = order>>8;
+unsigned inline int rans_compress_bound_4x16(unsigned int size, int order) {
```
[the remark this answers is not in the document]
jkbonfield: It's just a request to the compiler to inline it and perfectly norma and legal C.  Inline isn't limited purely to header files.  However it's not really necessary for most modern compilers as local functions tend to be decided on automatically.  So I'll just remove it.

htscodecs/rANS_static16_int.h
jkbonfield: It's definitely not something the code should be changing.  I'll amend the spec.

daviesrob:
There's an odd build failure on AppVeyor here:
```
C:\msys64\tmp\ccSsDBiO.s:6447: Error: invalid register for .seh_savexmm
```

[This gcc bug report](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65782) suggests it should be fixed on newer versions of gcc, but the one you're getting from AppVeyor seems to be quite old (mingw-w64-x86_64-gcc-9.1.0-3).  The AppVeyor docs [say](https://www.appveyor.com/docs/build-environment/#choosing-image-for-your-builds) that you get a `Visual Studio 2015` image if you don't ask for anything else (their images are named after Visual Studio releases).  Adding:
```
image: Visual Studio 2022
```
to `.appveyor.yml` will hopefully make the problem go away.  (Although note that HTSlib currently uses 2019 - it would be useful to know if that works as well).

daviesrob:
I've pushed a commit to make the AppVeyor build work.  It [shows up a problem](https://ci.appveyor.com/project/samtools/htscodecs/builds/42774591) in the entropy test.  I suspect this is due to high stack usage in some of the new order-1 rANS code.  E.g., `rans_uncompress_O1_4x16` looks like it uses over 1Mb of stack, which I understand may be a problem on Windows.

jkbonfield:
I fixed something related to that recently, so can apply a similar fix.

Ironically it was in wanting to bug test that change and to ensure my pthread_once tokens don't trample each other that I produced the new test binary.  I those pthread_once keys possibly were clashing for the older code, but I still need to explore that by mixing non-AVX and AVX versions in the same tool.  More work to be done there probably.

jkbonfield:
My thought is if we have stack memory allocated as a global pthread_once buffer for function A, and similarly for function B, but it's the same pthread_once token then we may end up allocating in incorrectly sized buffer (it needs to be the maximum of A or B).  That needs careful thought as I'm not testing all permutations of ordering.

It's safer to simply have a different token so each time we're avoiding a large stack we allocate a different global buffer for that specific problem.    I can see multiple uses of `rans_once`, so I need to double check the sizing or deduplicate that name.

daviesrob:
I think the best solution to the per-thread memory allocations might be to wrap them in a simple API.  It would just need two functions, e.g. `mem_block = codecs_get_tls_mem(size);` and `codecs_release_tls_mem(mem_block);`.  Callers then just need to specify the size they want, and the get function can look after supplying a big enough block.

Another advantage of this is that it could wrap up the threads / no-threads variants, which would simplify the functions that use these allocations.  It could also be extended to handle nested functions that use these blocks, should the need arise.

jkbonfield:
Agreed.  It was fine as-is when it was a one off, but now the style is being copied multiple times it needs some sanitizing.

jkbonfield:
Checking this again I see it's far from obvious, but it is correct.  I needs tidying up though.

So rans_once / rans_key are used multiple times.  However *mostly* they're static so it's a "local" global variable and they're using file-specific copies.  This is important as the malloc buffer used by 4x16 order-1 is ~1MB while the 32x16 AVX versions are ~4MB.  The except is the 32x16 scalar implementation, which has:

```
// Reuse the rANS_static4x16pr variables                                        
extern pthread_once_t rans_once;                                                
extern pthread_key_t rans_key;                                                  
```

As per the comment, the rANS4x16pr.c has non-static versions of these variables and so they share the same pointer.  That means it ends up working fine.

Frankly a better bet is to simply avoid that reuse as it's only 1MB duplicated so hardly breaking the bank by modern systems.  Everything can then be static and no symbol leakage.

While having a dedicated tls allocation function could also work, it's tricky to get this done with arbitrary sizes.  We could do it, but we're basically writing a caching layer and having to reuse the same `pthead_once` key throughout with a search through known blocks looking for sizes.  I don't see the point of a free function either as it won't actually free anything.  (If it did then we're back to square one of just using malloc/free and the entire thing becomes pointless.)

daviesrob:
I don't think the allocation function would be that complicated.  It would just need to check if the block it has is big enough and replace it with a bigger one if not.  It would rapidly get to the largest size needed and then stop growing.  If you also kept track of in-use status, you'd also be able to catch bugs like forgetting to release the block, or calling a function that also tries to use the same block.

jkbonfield:
It's passing Windows MINGW tests at last.  It would be good for someone to test it by using this new branch via Scramble on a Mac to validate multi-threading, which also exhibits a *pitifully* small stack per thread to see if there are more places we need to make use of TLS.

daviesrob:
I've tried it on my Mac.  The latest adjustments worked, but so also did the previous commit because the threads are [created with a 4Mb stack](https://github.com/jkbonfield/io_lib/blob/993bbda03ffb4f6e66a5beedcbbddf11637f4da2/io_lib/thread_pool.c#L481).  Unfortunately if I remove the stack-setting attribute I still get a couple of test failures, even with the new allocation function:

```
../progs/scramble -t4 -V4.0 -7 -r ./data/c1.fa test.out/c1#bounds.bam test.out/c1#bounds.full.cram

WARNING: this version of CRAM is not a recognised GA4GH standard.
Note this CRAM version is a technology demonstration only.
Future versions of Scramble may not be able to read these files.

Bus error: 10
```

I'll see if I can work out what's going on.

daviesrob:
Aha! `rans_compress_O1_4x16` still [tries to use a lot of stack](https://github.com/samtools/htscodecs/blob/d7472068/htscodecs/rANS_static4x16pr.c#L421).

jkbonfield:
The same code is in `rans_compress_O1_32x16` too.  I'll work on an extended fix.

daviesrob:
I'm still getting problems with high stack usage.  A quick `grep -r '\[256]\[' .` suggests there may still be problems in [`rans_uncompress_O1_4x16`](https://github.com/samtools/htscodecs/blob/95d5373e/htscodecs/rANS_static4x16pr.c#L543),  [`rans_uncompress_O1_32x16`](https://github.com/samtools/htscodecs/blob/95d5373e/htscodecs/rANS_static32x16pr.c#L524) [(also line 528)](https://github.com/samtools/htscodecs/blob/95d5373e/htscodecs/rANS_static32x16pr.c#L528), [`rans_uncompress_O1_32x16_neon`](https://github.com/samtools/htscodecs/blob/95d5373e/htscodecs/rANS_static32x16pr_neon.c#L1390) [(also line 1394)](https://github.com/samtools/htscodecs/blob/95d5373e/htscodecs/rANS_static32x16pr_neon.c#L1394), and [`encode_freq1`](https://github.com/samtools/htscodecs/blob/95d5373e/htscodecs/rANS_static16_int.h#L319).

jkbonfield:
Meh...  I'm starting to wonder how useful this avenue is!  I can try and see what impact it has on speed though.   Now we have the TLS alloc function it's pretty straight forward to replace them.

htscodecs/rANS_static.c
```diff
@@ -682,28 +570,13 @@ unsigned char *rans_uncompress_O1(unsigned char *in, unsigned int in_size,
...
-    ari_decoder *const D = calloc(256, sizeof(*D));
-    RansDecSymbol32 (*const syms)[256] = malloc(256 * sizeof(*syms));
-    for (i = 1; i < 256; i++) memset(&syms[i][0], 0, sizeof(syms[0][0]));
-#else
-    ari_decoder D[256] = {{{0}}}; //256*4k    => 1.0Mb
-    RansDecSymbol32 syms[256][256+6] = {{{0}}}; //256*262*8 => 0.5Mb
-#endif
-#endif
+    RansDecSymbol32 (*syms)[256];
+    uint8_t *mem = htscodecs_tls_calloc(256, sizeof(ari_decoder)
```
daviesrob, line 574: There are a number of `return NULL;` statements below that could cause this memory to leak.  They either need to be changed to `goto cleanup;` or the checks involved should be moved above this allocation.

htscodecs/rANS_static32x16pr.c
```diff
@@ -0,0 +1,714 @@
...
+    if (!out || bound > *out_size)
+	return NULL;
+
+    if (((size_t)out)&1)
+	bound--;
+    out_end = out + bound;
+
+    RansEncSymbol (*syms)[256] = htscodecs_tls_alloc(256 * (sizeof(*syms)));
+    if (!syms)
+	return NULL;
```
daviesrob, line 415: `out` could be leaked here (if it was allocated above).

htscodecs/rANS_static32x16pr.c
```diff
@@ -0,0 +1,714 @@
...
+
+    cp = out;
+    int shift = encode_freq1(in, in_size, 32, syms, &cp); 
+    if (shift < 0) {
+	htscodecs_tls_free(syms);
+	return NULL;
```
daviesrob, line 421: `out` could also be leaked here.

htscodecs/rANS_static4x16pr.c
```diff
@@ -1147,18 +1080,26 @@ unsigned char *rans_compress_to_4x16(unsigned char *in,  unsigned int in_size,
 	if (!(out = malloc(*out_size)))
 	    return NULL;
     }
+    if (*out_size == 0)
+	return NULL;
```
daviesrob, line 1084: This test should happen before trying to allocate `out`.  Also, `out` could be leaked later if allocated above and one of a number of other allocations fails.

daviesrob (commented):
Similar leakages can also occur in the SIMD versions of these functions, as they include similar checks to the non-SIMD ones.

htscodecs/rANS_static32x16pr_avx2.c
```diff
@@ -0,0 +1,1618 @@
...
+    int c[8] __attribute__((aligned(32)));
+    _mm256_store_si256((__m256i *)c, idx);
+    return _mm256_set_epi32(b[c[7]], b[c[6]], b[c[5]], b[c[4]],
+			    b[c[3]], b[c[2]], b[c[1]], b[c[0]]);
+}
+#else
+#define _mm256_i32gather_epi32x _mm256_i32gather_epi32
+#endif
+
+static inline __m128i _mm_i32gather_epi32x(int *b, __m128i idx, int size) {
```
daviesrob, line 136: The compiler complains that this is unused.

tests/entropy.c
```diff
@@ -0,0 +1,189 @@
...
+		free(comp);
+		free(uncomp);
+	    }
+	}
+	printf("\n");
+    }
+
+    free(in);
+
+    return 0;
```
daviesrob, line 188: This program can't return a failure code to the test harness.

I have an improved version that can, and also check that the compressed data is the same between the alternate optimised versions of the algorithms.  Would you like it?

jkbonfield: Yes that's a bit of an oversight.  Thank you.

I have an amendment to it also to set stack size so we can validate it passes on e.g. Alpine linux. (It does, to be pushed as soon as I bug fix one other case.)

daviesrob: Calling `pthread_exit()` also means you can't return an exit code.  Is it there simply so you can check that there is no reachable memory left using `valgrind`?  Neither `valgrind` nor Address Sanitizer seem too worried about the memory that gets left behind without doing this.

It's a bit annoying that the destructor passed to `pthread_key_create()` doesn't get called when the main thread exits, though.  It might be possible to work around by registering an `atexit()` handler, but it's a bit difficult to tell if the result would be to free the memory, or to go catastrophically wrong...

htscodecs/rANS_static4x16pr.c
```diff
@@ -540,7 +542,8 @@ unsigned char *rans_uncompress_O1_4x16(unsigned char *in, unsigned int in_size,
 
     if (!sfb_)
 	return NULL;
-    fb_t fb[256][256];
+    //fb_t fb[256][256];
+    fb_t (*fb)[256] = htscodecs_tls_alloc(256 * sizeof(*fb));
```
daviesrob, line 546: This allocation also needs to be checked for failure; or it could be combined with the one for `sfb_` as done in `rans_uncompress_O1_32x16()`.

daviesrob:
I've tried a bit of fuzz testing on the new codecs.  So far the main discovery has been an endless loop, reached by feeding these bytes into `rans_uncompress_4x16()`:

```
00000000  2b 3d 00 3d                                       |+=.=|
```

This gets it into `RANS_ORDER_STRIPE` mode, with a stripe size of 0.  It gets stuck in `unstripe()` because it can't make any progress when `N == 0`.

daviesrob:
Also, UBSAN complains quite a lot about:
```
utils.h:208:35: runtime error: passing zero to clz(), which is not a valid argument
```
A simple fix would be:
```
e -= F0[i] * (32 - __builtin_clz(F0[i]|1) + in_size_r2);
```

I've also seen
```
rANS_static32x16pr_avx512.c:134:58: runtime error: left shift of negative value -32
```
but haven't had a chance to look into that one yet.

daviesrob:
I've tracked down the erroneous left-shift now.  It's actually a bit more naughty than it at first looks, as it's really using an uninitialised value.  It comes about because `syms` array entries are only set for [symbols that are actually in use](https://github.com/samtools/htscodecs/blob/ca19da75/htscodecs/rANS_static32x16pr_avx512.c#L116), but the loop that copies them into `SA, SB, SC and SD` [goes through everything](https://github.com/samtools/htscodecs/blob/ca19da75/htscodecs/rANS_static32x16pr_avx512.c#L131).  So for unused symbols it will get random data from the stack, although it shouldn't matter too much as the bad data will be ignored after this point.

A simple solution is to make the loop that fills out `syms` clear out the unused entries, and set `rcp_shift` to something >= 32 to avoid the UB later.

I have a couple of tiny UB fixes now.  Would you like me to add them as a commit?

jkbonfield:
> I have a couple of tiny UB fixes now. Would you like me to add them as a commit?

Provided it doesn't unduly slow things down by zeroing memory that's never going to have any impact on code execution.  I have a benchmark script in the test directory.  Just give it a file to work on.

daviesrob:
This could do with a bit of a squash, prior to being merged.  Would you like to do the honours?

jkbonfield:
I'll have a play. :)

jkbonfield:
Squashed as much as I desire.  I haven't squashed things between authors.  I reordered a few minor things to squash, and got most of the hts_tls_alloc bits into the same commit amongst other groupings.

htscodecs/rANS_static4x16pr.c
```diff
@@ -1114,24 +746,329 @@ unsigned char *rans_uncompress_O1_4x16(unsigned char *in, unsigned int in_size,
...
+
+// We may well be able to write generate AVX2 code, but if we can't auto-detect
+// it then it's pointless.  We can however still use the 32-way codec as
+// we may be decoding on a machine with AVX2 support and the CPU hit isn't
+// vast.
+#ifdef HAVE_AVX2
+#  undef HAVE_AVX2
+#endif
+
+#ifdef HAVE_AVX2
```
jmarshall, line 1032: I don't really understand what undefining these here achieves, but this one is surely a typo for `#ifdef HAVE_AVX512`.

jkbonfield: There were a few typos from code kill-and-yank operations of this nature, but I thought we'd corrected them all.

The undef seem to be a relic though so the entire thing could be culled.  When I initially started work on these I just had one source file that `#include`d the various others based on the existance of the definitions, so undefining them made sence.  Once I got further into the project I did it properly via Makefile rules and file-specific compiler opts (the only reliable way to work infact), but it looks like some left-overs weren't culled.

PR #27, opened 2021-07-05
Update to release 1.1

jkbonfield:
No code changes, just version numbering and docs.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -32,7 +32,7 @@
 AUTOMAKE_OPTIONS = foreign no-dependencies
 ACLOCAL_AMFLAGS = -I m4
 
-SUBDIRS = htscodecs tests
+SUBDIRS = htscodecs . tests
 
 nobase_include_HEADERS = \
 	htscodecs/arith_dynamic.h \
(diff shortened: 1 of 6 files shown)


jmarshall:
You might also have a look at whatever patches Debian currently applies. See https://salsa.debian.org/med-team/htscodecs/-/tree/master/debian/patches — the only current one (as listed in _series_) is _link_with_libm_, which looks like it would be usefully applicable upstream.

jkbonfield:
Good catch, and gah!  Why do they never feed back information upstream?

jmarshall:
Why indeed… it's super frustrating :smile:

jkbonfield:
I snuck in the `-lm` commit prior to the news/version update here, as otherwise it's needless red tape and extra shuffling / rebasing when making the release.  This PR can act as validation it works via the tests (which it does for me locally).

daviesrob:
Updated with my improved distcheck fix.  The problem was that `tests/varint` was accidentally included in `EXTRA_DIST`, which caused make to attempt to build it earlier than it should have.

PR #30, opened 2021-11-02
Fix off-by-one error on initial size check for arith_dynamic.

jkbonfield:
This meant the most naive of small tests would fail to decode as the compressed data size exactly consumes the size of the initialisation of `rc->code`.  We incorrectly rejected this case.

Note this also shows a weakness of error reporting, as the caller needs a better way to be told of the input data being too small (which admittedly it wasn't, but the bug meant it thought it was but no report was given).

That's a much wider change, for another PR some time, as it'll need the error handling in c_simple_model.h completely redesigning.


daviesrob:
Fix looks OK, although it might be good to have a comment on why the first byte of data is always discarded in the code that follows...

PR #31, opened 2021-11-16
Various rANS speedups

jkbonfield:
A mixture of bits cherry-picked out of the unroll32 branch.  These apply to the main codecs and not just the 32-way SIMD implementation.  Some perf stat reports from 10 runs on low complexity (q4) and high complexity (q40) data sets comparing master (old) vs this PR (new).

Clang10
```
                    Giga-cycles enc     Giga-cycles dec
                    old      new        old      new
4x8 -o1 q4          2186     2061 -     1225     1202 -
4x8 -o1 q40         1570     1482 -     1033     1034 ~
4x16 -o193 q4       1533     1435 -      778      762 -
4x16 -o1 q40        2553     1789 --     863      868 ~
```
Gcc11
```
                    Giga-cycles enc     Giga-cycles dec
                    old      new        old      new
4x8 -o1 q4          2174     2075 -     1234     1205 -
4x8 -o1 q40         1667     1574 -     1045     1045 =
4x16 -o193 q4       1694     1573 -      775      789 +
4x16 -o1 q40        2539     1798 --     865      867 ~
```

New decoder is marginally different, sometimes better, sometimes worse, but varies by file / compiler. All minor and maybe just noise.  Nothing was specifically changed here, so any genuine speed differences are likely simply down to function alignment.

Encoder is universally faster, considerably so for 4x16 on complex data (q40).

### DIFF
--- a/htscodecs/utils.h
+++ b/htscodecs/utils.h
@@ -46,17 +46,44 @@ static inline void unstripe(unsigned char *out, unsigned char *outN,
     if (ulen >= N) {
 	switch (N) {
 	case 4:
+#define LLN 16
+	    if (ulen >= 4*LLN) {
+		while (j < ulen-4*LLN) {
+		    int l;
+		    for (l = 0; l < LLN; l++) {
+			for (k = 0; k < 4; k++)
+			    out[j+k+l*4] = outN[idxN[k]+l];
+		    }
+		    for (k = 0; k < 4; k++)
+			idxN[k] += LLN;
+		    j += 4*LLN;
+		}
+	    }
 	    while (j < ulen-4) {
 		for (k = 0; k < 4; k++)
 		    out[j++] = outN[idxN[k]++];
 	    }
+#undef LLN
 	    break;
 
 	case 2:
+#define LLN 4
+	    if (ulen >= 2*LLN) {
+		while (j < ulen-2*LLN) {
+		    int l;
+		    for (l = 0; l < LLN; l++) {
+			for (k = 0; k < 2; k++)
+			    out[j++] = outN[idxN[k]+l];
+		    }
+		    for (k = 0; k < 2; k++)
+			idxN[k] += l;
+		}
+	    }
 	    while (j < ulen-2) {
 		for (k = 0; k < 2; k++)
 		    out[j++] = outN[idxN[k]++];
 	    }
+#undef LLN
 	    break;
 
 	default:
--- a/htscodecs/utils.h
+++ b/htscodecs/utils.h
@@ -195,6 +195,7 @@ void hist1_4(unsigned char *in, unsigned int in_size,
 	    l = c;
 	}
     }
+    T0[l]++;
 
     int i;
     for (i = 0; i < 256; i++)
--- a/htscodecs/utils.h
+++ b/htscodecs/utils.h
@@ -79,26 +79,56 @@ static inline void unstripe(unsigned char *out, unsigned char *outN,
  */
 static inline
 void hist8(unsigned char *in, unsigned int in_size, uint32_t F0[256]) {
-    uint32_t F1[256+MAGIC] = {0}, F2[256+MAGIC] = {0}, F3[256+MAGIC] = {0};
-    uint32_t F4[256+MAGIC] = {0}, F5[256+MAGIC] = {0}, F6[256+MAGIC] = {0};
-    uint32_t F7[256+MAGIC] = {0};
+    if (in_size > 500000) {
+	uint32_t f0[65536+37] = {0};
+	uint32_t f1[65536+37] = {0};
+	uint32_t f2[65536+37] = {0};
 
-    unsigned int i, i8 = in_size & ~7;
-    for (i = 0; i < i8; i+=8) {
-	F0[in[i+0]]++;
-	F1[in[i+1]]++;
-	F2[in[i+2]]++;
-	F3[in[i+3]]++;
-	F4[in[i+4]]++;
-	F5[in[i+5]]++;
-	F6[in[i+6]]++;
-	F7[in[i+7]]++;
+	uint32_t i, i8 = in_size & ~15;
+
+	for (i = 0; i < i8; i+=16) {
+	    uint16_t i16a[4], i16b[4];
+	    memcpy(i16a, in+i, 8);
+	    f0[i16a[0]]++;
+	    f1[i16a[1]]++;
+	    f2[i16a[2]]++;
+	    f0[i16a[3]]++;
+
+	    memcpy(i16b, in+i+8, 8);
+	    f1[i16b[0]]++;
+	    f0[i16b[1]]++;
+	    f1[i16b[2]]++;
+	    f2[i16b[3]]++;
+	}
+
+	while (i < in_size)
+	    F0[in[i++]]++;
+
+	for (i = 0; i < 65536; i++) {
+	    F0[i & 0xff] += f0[i] + f1[i] + f2[i];
+	    F0[i >> 8  ] += f0[i] + f1[i] + f2[i];
+	}
+    } else {
+	uint32_t F1[256+MAGIC] = {0}, F2[256+MAGIC] = {0}, F3[256+MAGIC] = {0};
+	uint32_t i, i8 = in_size & ~7;
+
+	for (i = 0; i < i8; i+=8) {
+	    F0[in[i+0]]++;
+	    F1[in[i+1]]++;
+	    F2[in[i+2]]++;
+	    F3[in[i+3]]++;
+	    F0[in[i+4]]++;
+	    F1[in[i+5]]++;
+	    F2[in[i+6]]++;
+	    F3[in[i+7]]++;
+	}
+
+	while (i < in_size)
+	    F0[in[i++]]++;
+
+	for (i = 0; i < 256; i++)
+	    F0[i] += F1[i] + F2[i] + F3[i];
     }
-    while (i < in_size)
-	F0[in[i++]]++;
-
-    for (i = 0; i < 256; i++)
-	F0[i] += F1[i] + F2[i] + F3[i] + F4[i] + F5[i] + F6[i] + F7[i];
 }
 
 /*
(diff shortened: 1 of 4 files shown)


daviesrob:
The comment on a644558 makes it look like the new version is slower.  Hopefully that's not true...

The threshold used to choose the method for `hist8` looks like its too low.  At 500000, I find the 16-bit version runs at half the speed of the 8-bit one.  It's got to zero out and sum over 65536*3*4 = 786432 bytes of data while collecting the histogram, so it looks like it needs quite a bit of input to compensate for the overhead.  I find the break-even point is somewhere around 2Mbytes, with significant gains beyond that especially when the alphabet is small.  Here are some test results:

```
clang 10.0.1 -O3; 256 chars alphabet; Intel Core Processor (Broadwell, IBRS)
   500000: 8-bit 0.000301740 16-bit 0.000944543
  1000000: 8-bit 0.000535523 16-bit 0.001113932
  2000000: 8-bit 0.001084657 16-bit 0.001323339
  5000000: 8-bit 0.005861814 16-bit 0.003489994
 10000000: 8-bit 0.005317978 16-bit 0.004189553
 20000000: 8-bit 0.010442628 16-bit 0.008444177
 50000000: 8-bit 0.026469041 16-bit 0.020214469

clang 10.0.1 -O3; 7 chars alphabet; Intel Core Processor (Broadwell, IBRS)
   500000: 8-bit 0.000274182 16-bit 0.000777248
  1000000: 8-bit 0.000615437 16-bit 0.000935147
  2000000: 8-bit 0.001195033 16-bit 0.001283486
  5000000: 8-bit 0.002771298 16-bit 0.002005068
 10000000: 8-bit 0.005424819 16-bit 0.003396203
 20000000: 8-bit 0.011040170 16-bit 0.006003091
 50000000: 8-bit 0.028343387 16-bit 0.016670147
```

Trying an older machine gives similar results:
```
clang 10.0.1 -O3; 256 chars alphabet; Intel(R) Xeon(R) CPU E5-2660
   500000: 8-bit 0.000303757 16-bit 0.000691956
  1000000: 8-bit 0.000615331 16-bit 0.000891452
  2000000: 8-bit 0.001239247 16-bit 0.001244705
  5000000: 8-bit 0.003094876 16-bit 0.002407654
 10000000: 8-bit 0.006267579 16-bit 0.004652264
 20000000: 8-bit 0.012711485 16-bit 0.008320565
 50000000: 8-bit 0.032328439 16-bit 0.020378659

clang 10.0.1 -O3; 7 chars alphabet; Intel(R) Xeon(R) CPU E5-2660
   500000: 8-bit 0.000322132 16-bit 0.000664558
  1000000: 8-bit 0.000656585 16-bit 0.000804381
  2000000: 8-bit 0.001353914 16-bit 0.001144487
  5000000: 8-bit 0.003236168 16-bit 0.001998801
 10000000: 8-bit 0.006316081 16-bit 0.003643112
 20000000: 8-bit 0.013103202 16-bit 0.007068384
 50000000: 8-bit 0.032875647 16-bit 0.017594091
```

As does gcc:
```
gcc 11.1.0 -O3; 256 chars alphabet; Intel Core Processor (Broadwell, IBRS)
   500000: 8-bit 0.000296290 16-bit 0.000940208
  1000000: 8-bit 0.000584834 16-bit 0.000995095
  2000000: 8-bit 0.001021722 16-bit 0.001407867
  5000000: 8-bit 0.002770676 16-bit 0.002461731
 10000000: 8-bit 0.005563724 16-bit 0.004253281
 20000000: 8-bit 0.010529946 16-bit 0.008131470
 50000000: 8-bit 0.028000319 16-bit 0.021223718

gcc 11.1.0 -O3; 7 chars alphabet; Intel Core Processor (Broadwell, IBRS)
   500000: 8-bit 0.000260284 16-bit 0.000906153
  1000000: 8-bit 0.000606467 16-bit 0.000922202
  2000000: 8-bit 0.001116496 16-bit 0.001250470
  5000000: 8-bit 0.002735532 16-bit 0.002286767
 10000000: 8-bit 0.005574476 16-bit 0.003525794
 20000000: 8-bit 0.011152284 16-bit 0.006097752
 50000000: 8-bit 0.028374063 16-bit 0.015769319
```
I've attached the test program I used to get these figures in case you'd like to try reproducing them.

[ht.c.gz](https://github.com/samtools/htscodecs/files/7604019/ht.c.gz)

jkbonfield:
Hmm interesting.  I chose that number because I did loads of benchmarks and that's where the break even point came out.

I wonder what's changed?  I'll have to check more then.

jkbonfield:
There's definitely something different between the test harness and testing in-situ, but I haven't worked out what.  Maybe it's due to the static inline nature?  Regardless editing the code between `if (in_size > 500000 || 1) {` vs `if (in_size > 500000 && 0) {` to be on/off shows this:

on:
```
~/io_lib/htscodecs/build/tests/rans4x16pr -o4 -t ~/scratch/data/q40
Testing 96 blocks
790.2 MB/s enc, 2110.4 MB/s dec	 101000000 bytes -> 51143511 bytes
800.2 MB/s enc, 2112.9 MB/s dec	 101000000 bytes -> 51143511 bytes
...
```

off:
```
~/io_lib/htscodecs/build/tests/rans4x16pr -o4 -t ~/scratch/data/q40
Testing 96 blocks
743.3 MB/s enc, 2122.5 MB/s dec	 101000000 bytes -> 51143511 bytes
743.0 MB/s enc, 2117.8 MB/s dec	 101000000 bytes -> 51143511 bytes
...
```

That's system gcc, and running on a modern system.  On an ancient Opteron it's identical in timing, despite the more synthetic benchmark being VERY different (not liking the new version at all).  So the code does speed it up, but it needs more research to understand why.  Maybe it's some code alignment thing, inlining changing the register allocation, or simply due to repeated reuse changing cache timings.  It's possible my own test harness is wrong too...

jkbonfield:
I think some of this is due to predictability of the data.  Your synthetic benchmark is using cyclic values; just `i & mask` for increasing i.  My benchmark is loading an entire block of Illumina quality values.  It gives quite different results, most likely due to differences in cache misses.

Also your test is very quick.  If I run it in a loop multiple times then it starts to favour the new algorithm at lower values.  This is possibly the benefit of branch prediction kicking in?  I'm unsure.

Either way, maybe 500k is still too low.  It looks closer to 1Mb now.

See ~jkb/tmp/ht.c and ~jkb/tmp/ht2.c (loads a file from stdin) for improved metrics.

daviesrob:
Your test harness run was 101 Mb, so not too surprising that it favours the 16-bit version.  Having said that, it does start winning earlier when running inside `rans_compress_O0_4x16`.  In my tests 500000 is about the same, but it starts winning at about 750000.  Comparing `perf` outputs suggests that the difference comes from the `hist8()` part.  It's not entirely clear why that should happen, although I notice that the first run of the 16-bit version always seems to be a bit slower, so maybe it's down to the time needed to page in the stack space it uses?

```
Off:
250.2 MB/s enc, 413.6 MB/s dec	 750000 bytes -> 366594 bytes
248.1 MB/s enc, 422.4 MB/s dec	 750000 bytes -> 366594 bytes
249.0 MB/s enc, 351.7 MB/s dec	 750000 bytes -> 366594 bytes
249.2 MB/s enc, 406.1 MB/s dec	 750000 bytes -> 366594 bytes
251.8 MB/s enc, 414.3 MB/s dec	 750000 bytes -> 366594 bytes
249.2 MB/s enc, 422.6 MB/s dec	 750000 bytes -> 366594 bytes
249.4 MB/s enc, 422.6 MB/s dec	 750000 bytes -> 366594 bytes
248.7 MB/s enc, 411.2 MB/s dec	 750000 bytes -> 366594 bytes
251.6 MB/s enc, 415.4 MB/s dec	 750000 bytes -> 366594 bytes
248.8 MB/s enc, 422.5 MB/s dec	 750000 bytes -> 366594 bytes

On:
231.7 MB/s enc, 415.0 MB/s dec	 750000 bytes -> 366594 bytes
255.6 MB/s enc, 423.8 MB/s dec	 750000 bytes -> 366594 bytes
257.3 MB/s enc, 423.9 MB/s dec	 750000 bytes -> 366594 bytes
257.5 MB/s enc, 406.7 MB/s dec	 750000 bytes -> 366594 bytes
260.4 MB/s enc, 413.9 MB/s dec	 750000 bytes -> 366594 bytes
260.3 MB/s enc, 421.1 MB/s dec	 750000 bytes -> 366594 bytes
257.0 MB/s enc, 423.9 MB/s dec	 750000 bytes -> 366594 bytes
257.1 MB/s enc, 423.8 MB/s dec	 750000 bytes -> 366594 bytes
257.0 MB/s enc, 420.8 MB/s dec	 750000 bytes -> 366594 bytes
258.1 MB/s enc, 420.8 MB/s dec	 750000 bytes -> 366594 bytes
```

htscodecs/utils.h
```diff
@@ -46,17 +46,44 @@ static inline void unstripe(unsigned char *out, unsigned char *outN,
...
+	    }
 	    while (j < ulen-4) {
 		for (k = 0; k < 4; k++)
 		    out[j++] = outN[idxN[k]++];
 	    }
+#undef LLN
 	    break;
 
 	case 2:
+#define LLN 4
```
daviesrob, line 70: Why the difference for `LLN`? I'd have thought you could get away with bigger values for `case 2` than `case 4` as the `k` loop is smaller?

jkbonfield:
Regarding the test/rans4x16pr program, note it uses approximately 1Mb buffers (which is equivalent to 10,000 reads of 100bp long, so ballpark the sort of size we may get from a default quality stream on an Illuimna run in CRAM).  I give it a big file so it can operate on lots of data and aggregate the stats over the entire data set, and not just the same buffer repeatedly which could give misleading results).  It's probably the most robust method for testing speed changes.

htscodecs/utils.h
```diff
@@ -46,17 +46,44 @@ static inline void unstripe(unsigned char *out, unsigned char *outN,
...
+	    if (ulen >= 2*LLN) {
+		while (j < ulen-2*LLN) {
+		    int l;
+		    for (l = 0; l < LLN; l++) {
+			for (k = 0; k < 2; k++)
+			    out[j++] = outN[idxN[k]+l];
```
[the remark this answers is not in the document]
jkbonfield: To be honest I can't remember now!  I tuned case 2 and case 4 separately, trying different code styles for each to work out which worked.  I recall they weren't entirely consistent because of that, but whether this specific change was an accident or not I forget.

htscodecs/utils.h
jkbonfield: Because it's what worked best basically.  They were both manually tuned with a variety of platforms and compilers to pick the figures that worked for that data type.

daviesrob:
I think the `hist8()` difference was down to allocating the extra bit of stack on the first pass through - so if I adjusted my test program to do a dummy run before the actual timing one the new version would look a bit better.  Anyway, it looks like it should still generated correct outputs (or at least the same as it used to), so I have merged this.

PR #40, opened 2022-03-30
Assorted bug-fixes

daviesrob:
* Use consistent data types in `rans_compress_bound_4x16()`
* Reject stripe mode inputs with zero stripes, preventing an endless loop in `unstripe()`
* Fix UBSAN warnings about operations on NULL pointers, decoding bit-packed or RLE data.
* Reject attempts to duplicate non-existent name tokeniser streams.
* Fix possible use of uninitialised memory in rans_uncompress_O1_4x16().

The first three commits have been pulled over from #26, as they fix pre-existing bugs unrelated to the other updates there.

### DIFF
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -362,7 +362,7 @@ unsigned int rans_compress_bound_4x16(unsigned int size, int order) {
     if (!N) N=4;
 
     order &= 0xff;
-    int sz = (order == 0
+    unsigned int sz = (order == 0
 	? 1.05*size + 257*3 + 4
 	: 1.05*size + 257*257*3 + 4 + 257*3+4) +
 	((order & X_PACK) ? 1 : 0) +
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -911,16 +911,18 @@ unsigned char *rans_uncompress_O1_4x16(unsigned char *in, unsigned int in_size,
 
     uint8_t *sfb_ = pthread_getspecific(rans_key);
     if (!sfb_) {
-	sfb_ = calloc(256*(TOTFREQ_O1+MAGIC2), sizeof(*sfb_));
+	sfb_ = calloc(256*(TOTFREQ_O1+MAGIC2)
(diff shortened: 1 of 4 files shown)


htscodecs/rANS_static4x16pr.c
```diff
@@ -911,16 +911,18 @@ unsigned char *rans_uncompress_O1_4x16(unsigned char *in, unsigned int in_size,
...
 #else
-    uint8_t *sfb_ = calloc(256*(TOTFREQ_O1+MAGIC2), sizeof(*sfb_));
+    uint8_t *sfb_ = calloc(256*(TOTFREQ_O1+MAGIC2)
+                           + 256*256*sizeof(fb_t), sizeof(*sfb_));
 #endif
 
     if (!sfb_)
 	return NULL;
-    fb_t fb[256][256];
+    fb_t (*fb)[256] = (fb_t (*)[256]) (sfb_ + 256*(TOTFREQ_O1+MAGIC2));
```
jkbonfield, line 925: Can we not simply merge unroll32 already which has completely rewritten this?  You're just doing work that's already been done, and it'll cause conflicts later on.

I'm happy that unroll32 is sufficient now as it doesn't rapidly fail any fuzz testing, and anything new can appear in a subsequent PR anyway.  I don't really see the point of duplicating our work and causing us headaches later on in merging.

htscodecs/rANS_static4x16pr.c
```diff
@@ -1607,7 +1607,6 @@ unsigned char *rans_uncompress_to_4x16(unsigned char *in,  unsigned int in_size,
 		goto err;
 	}
     } else {
-	tmp1 = NULL;
```
jkbonfield, line 1608: Not that I disagree with the change (and have already accepted it in unroll32 branch), but I'm curious what's actually undefined here?

The C standard I checked states for things like memcpy that reading beyond the bounds of an array is (reasonably) undefined.  It doesn't state that a NULL pointer is undefined.  This feels a bit like code obscurity in order to silence a false positive.  Generally it's considered good behaviour to reset a pointer to NULL when we know we shouldn't be accessing it.  So removing this may be best done with a coment.

PR #43, opened 2022-04-12
Ensure CPU dispatch only references functions that exist

daviesrob:
Found while adding unroll_32 support to HTSlib.  Depending on which SIMD-related `HAVE_*` defines are present, some of the SIMD functions may not be compiled.  This caused build failures because the CPU dispatch functions referred to all of them.  This reworks the dispatchers so that they only call functions if they exist and have CPU support.
 
Also fixes missing indentation on 'if (do_simd) { ... }' by inverting the check and handling the false case at the top.

### DIFF
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -836,7 +836,11 @@ unsigned char *(*rans_enc_func(int do_simd, int order))
      unsigned int in_size,
      unsigned char *out,
      unsigned int *out_size) {
-    if (do_simd) {
+    if (!do_simd) { // SIMD disabled
+        return order & 1
+            ? rans_compress_O1_4x16
+            : rans_compress_O0_4x16;
+    }
 	unsigned int eax = 0, ebx = 0, ecx = 0, edx = 0;
 	int have_ssse3   = 0;
 	int have_sse4_1  = 0;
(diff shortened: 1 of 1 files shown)


jkbonfield:
I don't see any broken indentation in the old code.  I just checked it again.

Scratching my head for a while, given the level of testing this went through, I think the bug you're fixing is support for old compilers? I can see that would indeed break, and this fixes it (thanks).  Something I didn't test recently I guess, although it was working when initially written.  Looks like it got broke when refactoring.

jkbonfield:
Duh, should have done more testing.  It now produces warnings when you have a compiler that doesn't support these features.  So not fully fixed.  I'll add that to my own tidyup PR that's incoming.

jmarshall:
I ran into these build failures yesterday when I updated and ran `make` without rerunning configure, so didn't have `HAVE_AVX2` etc defined in _config.h_.

Then I distcleaned and reran configure, and the first thing I did was hack _config.h_ to pretend my compiler didn't support building AVX code to see what would happen. :bomb:

So :+1: to fixes along these lines.

jkbonfield:
Oddly enough that's exactly what I'm doing now with config.h.  This PR fixes it, but it needs attributes to avoid the warnings.  I'm testing how stable they are on other compilers atm (pragmas to disable warnings proved to be a royal PITA due to compilers masquerading as each other).

The "correct" way of more ifdefs works, and I had a version doing that, but it's totally ugly as they're *everywhere*!

PR #44, opened 2022-04-13
SIMD tidy up

jkbonfield:
Removed a warning introduced in #43.  Also fixed broken indentation that appeared then too.

A subsequent PR will land to change everything to spaces, to avoid non-standard tab configurations leading people up the garden path.  (To my mind, the only valid definition of a tab is the one used as standard by unix terminals, as that permits people to view source outside of a particular editor.)

### DIFF
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -854,7 +854,7 @@ unsigned char *(*rans_enc_func(int do_simd, int order))
             : rans_compress_O0_4x16;
     }
     unsigned int eax = 0, ebx = 0, ecx = 0, edx = 0;
-    // These may be unused, defending on HAVE_* config.h macros
+    // These may be unused, depending on HAVE_* config.h macros
     int have_ssse3   UNUSED = 0;
     int have_sse4_1  UNUSED = 0;
     int have_popcnt  UNUSED = 0;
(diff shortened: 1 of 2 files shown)


htscodecs/rANS_static4x16pr.c
```diff
@@ -841,44 +853,45 @@ unsigned char *(*rans_enc_func(int do_simd, int order))
...
-	int have_sse4_1  = 0;
-	int have_popcnt  = 0;
-	int have_avx2    = 0;
-	int have_avx512f = 0;
-
-	int level = __get_cpuid_max(0, NULL);
-	if (level >= 1) {
-	    __cpuid_count(1, 0, eax, ebx, ecx, edx);
+    unsigned int eax = 0, ebx = 0, ecx = 0, edx = 0;
+    // These may be unused, defending on HAVE_* config.h macros
```
daviesrob, line 857: defending?

daviesrob, line 857: defending?

htscodecs/rANS_static4x16pr.c
```diff
@@ -921,75 +934,76 @@ unsigned char *(*rans_dec_func(int do_simd, int order))
...
```
daviesrob, line 938: also defending here...

PR #47, opened 2022-04-22
Substantial reduction to name tokeniser memory.

jkbonfield:
We allocate last_context based on the current number of tokens used rather than MAX_TOKENS, although temporarily it is at this size and realloced down at the end.  There is maybe some further gain to be had with avoiding many mallocs and using a pooled allocator, but it's a little tricky with the realloc down.  Maybe slicing out of a single block of 100*128 tok and setting pointer into that, changing the end point down after parsing, and starting a new block when we get within 128 of the end (so max 1% wasted).  This is the simpler solution for now though.

The trie_t structure has also been rejigged to be 24 bytes instead of 32 bytes.  This is possible as MAX_NAMES is ~20 bits so we can use bit-fields.

(I have this currently undergoing fuzzing, but it hasn't blown up in the first couple minutes which is a start.)


htscodecs/tokenise_name3.c
```diff
@@ -122,19 +122,22 @@ enum name_type {N_ERR = -1, N_TYPE = 0, N_ALPHA, N_CHAR, N_DIGITS0, N_DZLEN, N_D
                 N_DIGITS, N_DDELTA, N_DDELTA0, N_MATCH, N_NOP, N_END, N_ALL};
 
 typedef struct trie {
-    char c;
-    int count;
     struct trie *next, *sibling;
-    int n; // Nth line
+    int count;
+    char c;
```
daviesrob, line 127: This won't be saving as much space as you'd like due to padding after `char c`.  Making it part of the bitfield should work though.

```suggestion
    uint32_t c:8;
```

jkbonfield: Hmm, I thought I'd done that.  Maybe I forgot to push it.  It's what the "The trie_t structure has also been rejigged to be 24 bytes instead of 32 bytes. This is possible as MAX_NAMES is ~20 bits so we can use bit-fields" is referring to.  I'll check again.

jkbonfield: Ah sorry, yes I see that below.  The comment interrupted it and I failed to notice the commit does indeed already include the bitfield as intended.

I could make the char c part of the bitfield as you suggest, just to be explicit, but certainly on the platform I'm testing it's not necessary.

```
#include <stdio.h>                                                              
#include <stdint.h>                                                             
                                                                                
typedef struct trie1 {                                                          
    struct trie1 *next, *sibling;                                               
    int count;                                                                  
    char c;                                                                     
    uint32_t n:24; // Nth line                                                  
} trie1_t;                                                                      
                                                                                
typedef struct trie2 {                                                          
    struct trie2 *next, *sibling;                                               
    int count;                                                                  
    uint32_t c:8;                                                               
    uint32_t n:24; // Nth line                                                  
} trie2_t;                                                                      
                                                                                
int main(void) {                                                                
    printf("%ld %ld\n", sizeof(trie1_t), sizeof(trie2_t));                      
    return 0;                                                                   
}                                                                               
```

This returns 24 for both.  It was 32 before, showing the gain has been achieved.

PR #51, opened 2022-05-13
Work around a bug found using gcc-12 -O2.

jkbonfield:
Reworking this avoids this issue with gcc-12 -O2

    In file included from rANS_static4x16pr.c:60:
    In function 'RansEncPutSymbol',
        inlined from 'rans_compress_O1_4x16' at rANS_static4x16pr.c:478:9:
    rANS_word.h:315:28: warning: 'ptr' may be used uninitialized [-Wmaybe-uninitialized]
      315 |     *pptr = (uint8_t *)(ptr-c);
          |                        ~~~~^~~
    rANS_static4x16pr.c: In function 'rans_compress_O1_4x16':
    rANS_static4x16pr.c:450:14: note: 'ptr' was declared here
      450 |     uint8_t* ptr = out_end;
          |              ^~~

At -O1 and -O3 it also goes away.  With rans4x16pr -o1 under gcc-12
-O2, not only do we get this warning it also gets incorrect code due
to an invalid x_max, overflowing the output buffer and crashing.

The cause boils down to strict aliasing rules.  char* (and I assume
uint8_t*?) is permitted to point to any data type, but under the
strict aliasing rules you cannot have e.g. float* and int* could
pointing to the same memory.  It's one way for the compiler to
optimise, as it can make more assumptions.

However, casting uint8_t** to uint16_t** falls under the aliasing ban
hammer as it's (equiv) char** not char*:

    uint16_t* ptr = *(uint16_t **)pptr;

Deref first (to uint8_t*) then cast is the valid alternative as the
thing being cast is now equivalent to a char*.

    uint16_t* ptr = (uint16_t *)*pptr;

This now works fine with gcc-12 -O2, but is included more for
information incase we find this in earlier versions or performance
issues with the newer code on other compilers.  However a memcpy and
keeping fully in 8-bit land seems better overall on gcc benchmarking
(and sometimes on clang).

### DIFF
--- a/htscodecs/rANS_word.h
+++ b/htscodecs/rANS_word.h
@@ -18,6 +18,7 @@
 
 #include <stdio.h>
 #include <stdint.h>
+#include <string.h>
 #include <assert.h>
 #include "htscodecs_endian.h"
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
Interesting.  I thought I built it with -Wstrict-aliasing=3 earlier to look for others, but maybe that was another compiler version.  Anyway yes I agree we should fix the others, especially the other rans_word renorm one which is basically an identical bug.  I'll do a new PR soon with further amendments.

Thanks

daviesrob:
`-Wstrict-aliasing=3` comes with `-Wall`, so I wouldn't expect it to show anything we don't usually find.  The [gcc man page](https://linux.die.net/man/1/gcc) says for level 1:
"Most aggressive, quick, least accurate. Possibly useful when higher levels do not warn but -fstrict-aliasing still breaks the code, as it has very few false negatives."

PR #55, opened 2022-07-11
Fix compression of data blocks close to 2GB in size.

jkbonfield:
(We still have a limit of >2GB though)

Reported by Divon Lan

### DIFF
--- a/tests/rANS_static_test.c
+++ b/tests/rANS_static_test.c
@@ -49,7 +49,7 @@
 #endif
 
 // Room to allow for expanded BLK_SIZE on worst case compression.
-#define BLK_SIZE2 ((int)(1.05*BLK_SIZE))
+#define BLK_SIZE2 (1.05*BLK_SIZE)
 
 // Max 4GB
 static unsigned char *load(FILE *infp, uint32_t *lenp) {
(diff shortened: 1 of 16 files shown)


tests/rANS_static_test.c
```diff
@@ -93,7 +93,7 @@ static unsigned char *load(FILE *infp, uint32_t *lenp) {
  */
 int main(int argc, char **argv) {
     int opt, order = 0;
-    unsigned char in_buf[BLK_SIZE2+257*257*3];
+    unsigned char *in_buf = malloc(BLK_SIZE2+257*257*3);
```
daviesrob, line 96: Address sanitizer complains that this gets leaked.

The similar changes to malloc'd blocks in the other tests seem to get a pass as the buffers being replaced are currently global.

jkbonfield: It's true, but I didn't think it'd matter given it's essentially the same as a huge static array (it's freed on program exit).  I forgot about automated tests highlighting it though.  I'll fix as it's useful to be able to use the test tools complaint-free to look for problems elsewhere.

I did also think about checking for NULL, but frankly if it can't fit then it wouldn't have fitted before and both would just die.  Not a problem given this isn't a production tool, unless a compiler is going to start asserting error checking.

daviesrob:
The fixes look OK.  I think it would be a good idea to add explicit checks on the 2 Gbyte limit though.  At the moment it's not obvious if there is a limit, and it looks like there will be problems if you try to go over that.

PR #59, opened 2022-08-02
V1.3.0 README / BENCHMARKS

jkbonfield:
Minor documentation improvements.


tests/entropy.c
```diff
@@ -175,6 +190,7 @@ int main(int argc, char **argv) {
                 }
 
                 // decode
+                gettimeofday(&tv2, NULL);
```
daviesrob, line 193: I think you need one of these straight after the compression part as well.  At the moment you're including the time to do some printing, and in the SIMD tests a memcmp(), which would make the compression times look worse than they really are.

PR #60, opened 2022-08-09
Fix bug with RANS_ORDER_STRIPE on large blocks.

jkbonfield:
The meta-data is 2 bytes for order + N, ? bytes for original size, and
`4*N` bytes for the N compressed sizes.  We forgot about the original
size so only left room for `2+5*N` (with 5 being the maximum size of a
32-bit var int).  So blocks larger than 2^28 risked not fitting in the
allocated meta-data size.

Reported by Divon Lan.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,5 @@
-Release 1.3.0: 19th July 2022
------------------------------
+Release 1.3.0: 9th August 2022
+------------------------------
 
 The primary change in this release is a new SIMD enabled rANS codec.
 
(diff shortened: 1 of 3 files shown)


daviesrob:
This extra space doesn't seem to be accounted for in the `compress_bound` functions.  In fact, `arith_compress_bound` doesn't account for `X_STRIPE` at all.  Does that matter?

daviesrob:
Ah, I see that's been fixed now.

jkbonfield:
> This extra space doesn't seem to be accounted for in the `compress_bound` functions. In fact, `arith_compress_bound` doesn't account for `X_STRIPE` at all. Does that matter?

Indeed I noticed that while proof reading my change.  Oops.  Well it's all incremental steps forward.  To be honest though the compress bound is a bit guessworky with an overly pessimistic 5% growth.  I've no idea what the theoretical worst case growth is as a proportion of the input size, but it's unlikely to be remotely close to that.

PR #64, opened 2022-10-06
Check pthread_once() return value and print diagnostic

jmarshall:
While building htscodecs on FreeBSD, I found that most of the tests were failing due to a segfault in `htscodecs_tls_free()` due to `tls` being NULL. After much debugging, it turned out that `pthread_once()` was not working and `htscodecs_tls_init()` was never called.

Checking the return code from `pthread_once()` and printing a diagnostic provides a chance to diagnose this situation quickly rather than just being presented with the eventual segfault.

It turns out that on FreeBSD if you neglect to link with `-lpthread` or `-lthr`, libc provides versions of the pthread functions including a [`pthread_once()` stub that just doesn't work](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=150959). So the cause of these test failures was neglecting to link with `‑lthr` or another threading library.

This was caused by htscodecs's _configure.ac_ check picking up the libc pthread stubs on FreeBSD:

```
dnl Checks for library functions.
AC_SEARCH_LIBS([pthread_join], [pthread])

…

checking for library containing pthread_join... none required
```

This can be worked around with `./configure LIBS=-lthr` or similar, but ideally _configure.ac_ would use `AX_PTHREAD` or some other more comprehensive pthread test.

### DIFF
--- a/htscodecs/rANS_static.c
+++ b/htscodecs/rANS_static.c
@@ -92,7 +92,10 @@ unsigned char *rans_compress_O0(unsigned char *in, unsigned int in_size,
     ptr = out_end = out_buf + (uint32_t)(1.05*in_size) + 257*257*3 + 9;
 
     // Compute statistics
-    hist8(in, in_size, (uint32_t *)F);
+    if (hist8(in, in_size, (uint32_t *)F) < 0) {
+        free(out_buf);
+        return NULL;
+    }
     tr = ((uint64_t)TOTFREQ<<31)/in_size + (1<<30)/in_size;
 
  normalise_harder:
@@ -404,7 +407,11 @@ unsigned char *rans_compress_O1(unsigned char *in, unsigned int in_size,
     out_end = out_buf + (uint32_t)(1.05*in_size) + 257*257*3 + 9;
     cp = out_buf+9;
 
-    hist1_4(in, in_size, (uint32_t (*)[256])F, (uint32_t *)T);
+    if (hist1_4(in, in_size, (uint32_t (*)[256])F, (uint32_t *)T) < 0) {
+        free(out_buf);
+        out_buf = NULL;
+        goto cleanup;
+    }
 
     F[0][in[1*(in_size>>2)]]++;
     F[0][in[2*(in_size>>2)]]++;
--- a/htscodecs/rANS_static16_int.h
+++ b/htscodecs/rANS_static16_int.h
@@ -322,7 +322,8 @@ static inline int encode_freq1(uint8_t *in, uint32_t in_size, int Nway,
         return -1;
     uint32_t T[256+MAGIC] = {0};
     int isz4 = in_size/Nway;
-    hist1_4(in, in_size, F, T);
+    if (hist1_4(in, in_size, F, T) < 0)
+        goto err;
     for (z = 1; z < Nway; z++)
         F[0][in[z*isz4]]++;
     T[0]+=Nway-1;
--- a/htscodecs/rANS_static32x16pr.c
+++ b/htscodecs/rANS_static32x16pr.c
@@ -96,7 +96,6 @@ unsigned char *rans_compress_O0_32x16(unsigned char *in,
     // Compute statistics
     double e = hist8e(in, in_size, F);
     int low_ent = e < 2;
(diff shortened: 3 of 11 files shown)


jkbonfield:
Thanks for diagnosing this (and apologies for delay in review).

Given it SEGV before, printing up an error message and continuing will aid debugging but will still lead to the crash.  Would it not be better to return NULL at that point so the caller can immediately bail out?

jmarshall:
There was quite a saga summarised inside that “After much debugging”… :smile:

The failure mode on FreeBSD is that `pthread_getspecific()` always returns NULL, so `htscodecs_tls_alloc()` always allocates a new arena (and then forgets about it) and `htscodecs_tls_free()` crashes because it's not expecting NULL. The latter could be altered so that it didn't crash, in which case everything would work fine except all the memory would be leaked.

It would be possible to change the code so that in this FreeBSD scenario, the routines decayed to `calloc`/`free` and no memory was leaked. However this would depend on `pthread_getspecific(uninitialised_key)` returning NULL, which the FreeBSD stub version does. But in general calling that without a properly set up key is undefined behaviour.

So it's not worth doing that. Having now looked at where `htscodecs_tls_alloc()` is called  and error checked, and its other possible failures, you're quite right and I'll add a `return NULL`.

jkbonfield:
Thanks John

jkbonfield:
I should also add that the entirety of that code has an `#ifdef NO_THREADS` guard which does cause it to degenerate down to malloc/free calls.  I think this may be enabled on MinGW builds if I recall, but I'm not certain.  No threading is perhaps a bit overly specific and maybe we need an alternative `|| defined(HTS_NO_TLS)` thing in there so we can explicitly disable it on some platforms via a more appropriate way.

It's worth pointing out though that this **is a considerable speed up**, particularly with the GNU C library which just loves to use mmap for malloc.  A consequence of large local buffers and mmap / munmap is that we obtain data from the system and then give it back repeatedly, which after a while starts thrashing in the kernel as it runs out of pre-zeroed pages.  We don't really care about zeroing data for security, as if we just kept the pages to start with things would go faster.  This is controllable of course via `mallopt`, but it's harsh to do that sort of thing in a library.  So it's best to get it working out of the box on freebsd I guess.

jmarshall:
This now also fails with `NULL` in the “can't happen” scenario when `pthread_once()` fails. All invocations of `htscodecs_tls_alloc()` were checking the result, but some invocations of `htscodecs_tls_calloc()` were not, so the second commit addresses that. And the third commit prevents a segfault in _test/entropy.c_. Both latter commits tested by intentionally breaking `htscodecs_tls_alloc()` to always return NULL.

This fixes the segfaults when the tests are built incorrectly on FreeBSD using their libc's stub pthreads routines. The multithreaded allocator is indeed a speed up, and FreeBSD benefits from it just fine — provided the user links against the good pthreads in `-lpthread` or `-lthr`.

This PR improves the effects, but htscodecs still needs to prevent the bad configuration: either with documentation (“on FreeBSD, use `./configure LIBS=-lthr`”) or by improving _configure.ac_'s pthreads test. I'd leave the direction and details of that bit up to you…

jkbonfield:
Merged as is currently, but I still need to think on how best to deal with the `-lthr` vs configure check.

We could attempt to do an AC_COMPILE and execute type test to validate that no only does the function exist, but that it also works, however that breaks cross-compilation.  (I've no idea if it works anyway - I may have already broken it)

The alternative is ditching `AC_SEARCH_LIBS` in favour of a `AC_CHECK_LIB` comboi, so we link against `pthread` if it exists and don't if it doesn't, meaning if we have a bunch of fake functions in the C library then they don't overrule the ones found in a genuine library.  That's pretty ugly, but doesn't have any cross-compilation problems.  I'll try that first and if it takes too long add a documation update.

jkbonfield:
I didn't notice this comment until I was well up to my neck in the autoconf manual again.

Anyway, I took an approach of AC_SEARCH_LIBS is nearly right, but the ordering is not (in this case).  The documented strategy there of a series of AC_LINK_ISELSE clauses is enough to explain how to solve it.  Hopefully!  If you already have a freebsd system to hand to test on it'll save me some time.

Thanks

PR #66, opened 2022-10-17
Update hts_pack to operate in line with the spec.

### LINKED ISSUE #65: Bit packing inputs with 256 symbols do not disable bit packing
Encoders that use `hts_pack` disable bit packing if there are more than 16 symbols, e.g., [rANS Nx16](https://github.com/samtools/htscodecs/blob/843d4f63b1c64905881b4648916a4d027baa1a1c/htscodecs/rANS_static4x16pr.c#L1266-L1271) and the [arithmetic range coder](https://github.com/samtools/htscodecs/blob/843d4f63b1c64905881b4648916a4d027baa1a1c/htscodecs/arith_dynamic.c#L844-L849). However, the checks that disable bit packing compare against [a potentially truncated value](https://github.com/samtools/htscodecs/blob/843d4f63b1c64905881b4648916a4d027baa1a1c/htscodecs/pack.c#L74). When attempting to bit pack inputs with 256 symbols, `out[c_meta_len]` is 0, `out[c_meta_len] > 16` fails, and the encoders do not disable bit packing.

An example using the rANS Nx16 test program (version 1.3.0):

```
$ ruby -e "print (0..255).to_a.pack('C*')" | ./rans4x16pr -r -o 128 | xxd
Took 46003 microseconds,   0.0 MB/s
00000000: a082 0000 8200 0001 0203 0405 0607 0809  ................
00000010: 0a0b 0c0d 0e0f 1011 1213 1415 1617 1819  ................
00000020: 1a1b 1c1d 1e1f 2021 2223 2425 2627 2829  ...... !"#$%&'()
00000030: 2a2b 2c2d 2e2f 3031 3233 3435 3637 3839  *+,-./0123456789
00000040: 3a3b 3c3d 3e3f 4041 4243 4445 4647 4849  :;<=>?@ABCDEFGHI
00000050: 4a4b 4c4d 4e4f 5051 5253 5455 5657 5859  JKLMNOPQRSTUVWXY
00000060: 5a5b 5c5d 5e5f 6061 6263 6465 6667 6869  Z[\]^_`abcdefghi
00000070: 6a6b 6c6d 6e6f 7071 7273 7475 7677 7879  jklmnopqrstuvwxy
00000080: 7a7b 7c7d 7e7f 8081 8283 8485 8687 8889  z{|}~...........
00000090: 8a8b 8c8d 8e8f 9091 9293 9495 9697 9899  ................
000000a0: 9a9b 9c9d 9e9f a0a1 a2a3 a4a5 a6a7 a8a9  ................
000000b0: aaab acad aeaf b0b1 b2b3 b4b5 b6b7 b8b9  ................
000000c0: babb bcbd bebf c0c1 c2c3 c4c5 c6c7 c8c9  ................
000000d0: cacb cccd cecf d0d1 d2d3 d4d5 d6d7 d8d9  ................
000000e0: dadb dcdd dedf e0e1 e2e3 e4e5 e6e7 e8e9  ................
000000f0: eaeb eced eeef f0f1 f2f3 f4f5 f6f7 f8f9  ................
00000100: fafb fcfd feff                           ......
```

jkbonfield:
Previously when too many symbols were present, we still returned packed data but with 1 byte per symbol.  It was then up to the caller to reject this as inappropriate use of pack.  This meant that hts_pack was a transparent transform that always gives back something even if it's actually just growing the data by 1 byte.

However the CRAMcodecs spec states this scenario is illegal.

The checks in rANS_static4x16pr.c were bugged for nsym==256 as there wasn't a check of "|| out[c_meta_len] == 0", but now it's a moot point as NULL will be returned.

Fixes #65

### DIFF
--- a/htscodecs/arith_dynamic.c
+++ b/htscodecs/arith_dynamic.c
@@ -842,7 +842,7 @@ unsigned char *arith_compress_to(unsigned char *in,  unsigned int in_size,
         int pmeta_len;
         uint64_t packed_len;
         packed = hts_pack(in, in_size, out+c_meta_len, &pmeta_len, &packed_len);
-        if (!packed || (pmeta_len == 1 && out[c_meta_len] > 16)) {
+        if (!packed) {
             out[0] &= ~X_PACK;
             do_pack = 0;
             free(packed);
--- a/htscodecs/pack.c
+++ b/htscodecs/pack.c
@@ -57,9 +57,6 @@ uint8_t *hts_pack(uint8_t *data, int64_t len,
                   uint8_t *out_meta, int *out_meta_len, uint64_t *out_len) {
     int p[256] = {0}, n;
     uint64_t i, j;
-    uint8_t *out = malloc(len+1);
-    if (!out)
-        return NULL;
 
     // count syms
     for (i = 0; i < len; i++)
@@ -75,13 +72,12 @@ uint8_t *hts_pack(uint8_t *data, int64_t len,
     j = n+1;
 
     // 1 value per byte
-    if (n > 16) {
-        *out_meta_len = 1;
-        // FIXME shortcut this by returning data and checking later.
-        memcpy(out, data, len);
-        *out_len = len;
-        return out;
-    }
+    if (n > 16)
+        return NULL;
+
+    uint8_t *out = malloc(len+1);
+    if (!out)
+        return NULL;
 
     // Work out how many values per byte to encode.
     int val_per_byte;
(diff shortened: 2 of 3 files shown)


daviesrob:
`hts_pack()` is also called in [arith_compress_to()](https://github.com/samtools/htscodecs/blob/7dc8927b8/htscodecs/arith_dynamic.c#L844), which needs the same change as done in `rans_compress_to_4x16()`.

Also, I note that the `return NULL` at the end of `hts_pack()` is unreachable, but that was also true before this PR.

jkbonfield:
Made the change to the arithmetic coder too.

The `return NULL`  in `hts_pack` is only non-reachable because `val_per_byte` is computed by a bunch of if-else clauses and is covered in the switch statement, but the compilers aren't smart enough to know this.  So the return is basically doing the same job as a `default:` label would do in silencing warnings.

PR #67, opened 2022-10-17
Fix a subtle inefficiency with CRAM 3.1 O1 freq encoding.

jkbonfield:
For order-1 we always encode a value Q in the context of the previous value P, so the frequency table needs to have an array for F[P][0..255]. Given the next value will be in the context of Q, this means we nearly always have F[P] and F[Q] arrays filled out.  However in the rare scenario that symbol Q appears once only and that is the very last byte, then F[Q][0..255] is a full array of zeros and doesn't need to be in the frequency table output.

In reality, it's largely irrelevant because this scenario is extremely rare and even when it is triggered it's totally harmless as it's just a case of storing additional meta-data that's never accessed.

However it came up in a code review, so here is a fix.

See https://github.com/samtools/hts-specs/pull/433#discussion_r989893592 and related discussion for more details.

### DIFF
--- a/htscodecs/rANS_static.c
+++ b/htscodecs/rANS_static.c
@@ -420,6 +420,12 @@ unsigned char *rans_compress_O1(unsigned char *in, unsigned int in_size,
         if (T[i] == 0)
             continue;
 
+        if (T[i] == 1 && in[in_size-1] == i) {
+            // last symbol can be ignored as it doesn't have following
+            // contexts.
+            continue;
+        }
+
         //uint64_t p = (TOTFREQ * TOTFREQ) / t;
         double p = ((double)TOTFREQ)/T[i];
     normalise_harder:


daviesrob:
I can see why this should be done, but doesn't it also need to happen in other places?  For example, [`encode_freq1()`](https://github.com/samtools/htscodecs/blob/7dc8927b8/htscodecs/rANS_static16_int.h#L374) for the rANS32x16 encoders, and the [rans](https://github.com/samtools/htscodecs/blob/7dc8927b8/javascript/rans.js#L448) and [rans4x16](https://github.com/samtools/htscodecs/blob/7dc8927b8/javascript/rans4x16.js#L921) versions of the Javascript code.

Also, looking at the Javascript version, I notice a subtle difference with the C version.  The JS versions calculate the total by [summing over `F[]`](https://github.com/samtools/htscodecs/blob/7dc8927b8/javascript/rans.js#L220) while the C versions all use [`T[i]`](https://github.com/samtools/htscodecs/blob/7dc8927b8/htscodecs/rANS_static16_int.h#L381) instead.  These are the same, except for the last character in the input which is counted in `T[]` but not `F[]`.  This has a couple of implications:
1. It looks like the JS  `NormaliseFrequencies0()` functions would do a divide by zero if the last character is unique, because `tot` would be zero in that case.
2. This means the C normalisation will be very slightly different for that character compared to the rest, even if it appears elsewhere.  Does this matter?

It does suggest an alternate solution, which would be to correct the `T[]` count for the last character when it's used in normalisation.  Then if the last character was unique the result would be zero and it would be skipped along with the other unused symbols; and if it appears elsewhere the normalisation will be exactly the same for this character as for the others.

jkbonfield:
Normalisation values are sort of irrelevant as it's up to the encoder to decide how those values add up.  When you have rounding errors, something needs increasing and something needs decreasing so the numbers add up, but there are many strategies and it doesn't make any difference to the decoder.  Indeed the whole essence of this PR being unnecessary is also because it doesn't make any difference to the encoder.  We're simply storing some extra unused data by adding an additional symbol with zero frequency (the default anyway).

I'll have to think if fudging T is a better solution.

jkbonfield:
Closing as although this PR is fine, the extra hoops of perfection everywhere else and the time taken to do it to get this PR through review isn't worth the minimal impact this change has.  It's just a bit of trivial tidying with no real impact.

PR #68, opened 2022-10-20
Improve checks for pthread calls on FreeBSD

jkbonfield:
Implemented an AX_SEARCH_LIBS_REV function as a minimal replacement of AC_SEARCH_LIBS but with the no-library case as the last thing to try. This solves the problem where we preferentially want an external library implementation over the built-in C library.

Addresses autoconf issue raised in #64

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -208,7 +208,7 @@ AC_CHECK_HEADERS(fcntl.h limits.h unistd.h malloc.h)
 AC_CHECK_HEADER(zlib.h)
 
 dnl Checks for library functions.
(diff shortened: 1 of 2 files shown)


jmarshall:
On my FreeBSD/arm64 VM, `./configure` produced

```
…
checking for function pthread_join... -lpthread
…
```

and `make` produced well-linked test executables that worked. However if `thr` is going to be listed, I guess ideally it would be selected on FreeBSD where it exists… (On this FreeBSD/arm64 machine, libpthread.so etc are just symlinks to libthr.*.)

jkbonfield:
Ok fair enough.  Trivial to switch them around.

PR #70, opened 2023-01-12
Add data stream introspection functions.

jkbonfield:
These are used by samtools cram_size subcommand to report information on the compression method being used along with some of the internal parameters fetched from the compressed data-stream.


jkbonfield:
Closing.  It's too complex to get these used within Samtools as it adds an extra layer of dependencies to resolve.  These small changes don't justify it.

So punted these functions to htslib instead, given I was only 60:40 sure on where they belonged anyway.

PR #72, opened 2023-02-01
Add arth+bzip2 (EXT flag) test file

jkbonfield:
This is from the Y coord in a block of novaseq read names, which demonstrates where bzip2 beats the other encoding methods.


daviesrob:
At the moment, if you don't have `libbz2`, `configure` will disable the code that uses it.  The resulting library will build, but this test will fail.

Given that disabling `bz2` could in the future make some files unreadable, should we make this more like HTSlib, where you have to deliberately disable `bz2` to turn it off?  So the default if it can't find the library would be to fail `configure` unless you say you really don't want it?

(Also, while testing this I spotted that htscodecs seems to link zlib, but it doesn't appear to be used anywhere.  Is that the case, and could the zlib parts of `configure` be disabled to remove the unneeded dependency?)

jkbonfield:
Agreed with making bz2 an failure unless we opt-out.  It won't be used in CRAM normally anyway unless going for the highest compression levels as we don't even enable arith mode until we get to the archive profile, which is where bzip2 starts being used too.  (At least I think not.)

PR #77, opened 2023-03-13
Guard SIMD tests with #ifdef tests for platforms

### LINKED ISSUE #76: SIMD code selection and configury is incorrect for multi-arch compilation
[This investigation has occurred in a pysam context, which uses htslib's bundled-htscodecs build system rather than htscodecs's native build system. So this issue might equally properly be raised in htslib, but I expect any fix will span both htscodecs and htslib anyway.]

It seems that on modern macOS the usual way to build universal object files and executables is to compile and link with `-arch x86_64 -arch arm64`. This runs the compiler twice, once for each target, and then runs `lipo` to combine the two temporary _.o_ files into a single multi-arch (‘universal’) _.o_ file.

This does not interact well with htslib's _configure.ac_ SIMD probes (and presumably something similar happens with htscodecs's _configure.ac_):

```
# CFLAGS=[…] -arch arm64 -arch x86_64
[…]
checking whether C compiler accepts -mssse3 -mpopcnt -msse4.1... no
checking whether C compiler accepts -mavx2... no
checking whether C compiler accepts -mavx512f... no
checking whether C compiler supports ARM Neon... no
```

Looking in _config.log_ we see that the underlying x86_64 compilation has accepted `-mssse3` etc but the underlying arm64 compilation has produced errors, so the overall compilation fails too — resulting in `no`. And vice versa for the Neon probe.

This results in _rANS_static32x16pr_neon.o_ et al being omitted from the build, but the `*_neon()` invocations in _rANS_static4x16pr.c_ are included independent of the configure answer. This in turn leads to linking failure or for shared objects a failure to find `rans_compress_O0_32x16_neon` at runtime, as encountered in pysam-developers/pysam#1149.

It would be simple enough to fix the `*_neon()` invocation guard to respect the `no` answer and avoid this crash. However this pessimizes execution time for this multi-arch build by omitting all the SIMD implementations — really for this universal ARM/X86-64 build the answer to all four checks should be `yes`, with the 3 x86 SIMD implementations being used for the underlying x86_64 build and the Neon implementation being used for the underlying ARM build.

daviesrob:
Required for MacOS multiarch binaries, where files get built for both x86 and ARM.  This means configure tests for compiler options need to pass even though they're being run for the wrong architecture, and source files need to include guards around the architecture-specific parts.

There is a risk that the configure SIMD tests will print "yes" when building for other than the target architectures.  However in most such cases the needed include files won't be present so the tests will work correctly.

Also make the NEON test define HAVE_NEON, so that the neon-specific code in rANS_static4x16pr.c can be excluded correctly if the test does happen to fail on an ARM-like platform.

Adjusts the MacOS CI test so that it makes a multiarch binary to ensure this all works.

Fixes #76

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -105,8 +105,14 @@ task:
   package_install_script:
     - brew install autoconf automake libtool
 
-  << : *COMPILE
+  # Try building and testing a multiarch library
+  compile_script:
+    - autoreconf -i
+    - ./configure CFLAGS="-g -O3 -Wall -Werror -arch arm64 -arch x86_64"
+    - make -j4
 
+  test_script:
+    - make check
 
 # # ----------
 # # Windows MINGW.
--- a/configure.ac
+++ b/configure.ac
@@ -130,9 +130,11 @@ AX_CHECK_COMPILE_FLAG([-mssse3], [
 ], [], [], [AC_LANG_PROGRAM([[
 	  #include "x86intrin.h"
 	]],[[
+	  #if defined(__x86_64__)
 	  __m128i a = _mm_set_epi32(1, 2, 3, 4), b = _mm_set_epi32(4, 3, 2, 1);
 	  __m128i c = _mm_shuffle_epi8(a, b);
 	  return *((char *) &c);
+	  #endif
 	]])
 ])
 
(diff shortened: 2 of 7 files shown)


daviesrob:
Hmm, not quite there yet....

PR #78, opened 2023-03-14
Make SIMD configure tests work for MacOS multiarch builds

### LINKED ISSUE #76: SIMD code selection and configury is incorrect for multi-arch compilation

daviesrob:
Multiarch builds source files for each target, and then merges the results together into a single universal object file.  This causes issues with configure tests that attempt to detect architecture-specific features because if they fail on one architecture then they effectively fail for all of them.  In htscodecs this caused x86_64 SIMD code to be disabled when trying to build an x86_64 / arm64 universal binary because the related configure checks failed on the arm64 side - and likewise the arm NEON code was disabled due to a failing x86_64 test.

To fix:

- Add `#ifdef __x86_64__` and `#ifdef __ARM_NEON` to the source code so that architecture-specific code is filtered out at compile time.
- Add `#ifdef __x86_64__` guards to the x86 SIMD configure checks so that they don't fail for multiarch builds.
- Refactor the x86 SIMD tests so that instead of checking if a specific compiler option works, they find out if it is actually needed to build the code.  They do this by attempting to compile first without and then (if that fails) with the option.  The result will either be `none` (no option needed), `unsupported` (compilation failed both with and without the option), or an option that needs to be used to enable the feature.
- Remove the NEON configure test, and compile `rANS_static32x16pr_neon.c` unconditionally.
- Add dummy symbols to some source files to avoid "empty translation unit" errors if the code they contain has all been skipped due to `#ifdef` checks.

The combination of the `#ifdef __x86_64__` guards and revised configure checks means that when building for `x86_64`, including as part of a multiarch build, the flags will be enabled and disabled as needed; while on other platforms the test code will be skipped and the configure check will determine that no special compiler flags are needed.

Fixes #76


jmarshall:
On the Intel side, what about `__i386__`?

jkbonfield:
We did discuss this briefly, but concluded the only i386 machines with SIMD support are basically modern CPUs running in 32-bit mode.   The code should work fine in 32-bit mode, but it won't gain access to the accelerated implementations.  I don't really know how big an issue this is, and whether we actually care given it's a rather unusual scenario.

Do you have a better feeling for the reasoning behind using new CPUs in old modes?

daviesrob:
I did think about adding `__i386__`, but then discovered that the accelerated code is [currently only called on x86_64](https://github.com/samtools/htscodecs/blob/cd0737fff5/htscodecs/rANS_static4x16pr.c#L829).  As James mentions, old CPUs won't have the accelerated instructions anyway so it was simpler to just use `__x86_64__` everywhere.

We could add `__i386__` if there's demand for it, but I think it would be best done as a separate project.

jmarshall:
I had an idea the popcnt one would work on real i386, but perhaps not and perhaps the other requirements for that implementation would disable it on i386 anyway.

Debian still build for i386 so I was expecting they would notice the difference. However in the light of that guard on the dispatcher just being `__x86_64__`, never mind.

jkbonfield:
Indeed, while popcnt itself may exist on older CPUs (I didn't check), we only use it in conjunction with a minimum of SSE4 I think.

I guess there is potential to use it elsewhere, but I'm not really minded to spend time optimising i386 :)

PR #85, opened 2023-05-17
Amd avx512

jkbonfield:
- Favour AVX2 over AVX512 for some rans combinations on AMD.  This may need changing again later, but this is a simple fix for now.

- Move cpu checking capabilities to a once-per-thread set of static globals, to avoid overhead.

- Added block-size adjustments to rans4x16pr test harness.  It's not there for other codecs, but could be easily added.

### DIFF
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -57,6 +57,10 @@
 #include <limits.h>
 #include <math.h>
 
+#ifndef NO_THREADS
+#include <pthread.h>
+#endif
+
 #include "rANS_word.h"
 #include "rANS_static4x16.h"
 #include "rANS_static16_int.h"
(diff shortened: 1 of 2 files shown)


htscodecs/rANS_static4x16pr.c
```diff
@@ -842,26 +846,28 @@ void rans_set_cpu(int opts) {
...
+#ifndef NO_THREADS
+static pthread_once_t rans_cpu_once = PTHREAD_ONCE_INIT;
+#endif
+
+static int have_ssse3   UNUSED = 0;
+static int have_sse4_1  UNUSED = 0;
+static int have_popcnt  UNUSED = 0;
+static int have_avx2    UNUSED = 0;
+static int have_avx512f UNUSED = 0;
+int is_amd              UNUSED = 0;
```
daviesrob, line 862: Should this be `static` as well as the others?

jkbonfield: Yes, I thought I'd done so.  Maybe I changed it briefly for testing and forgot to put it back.

Good spot.

Issue #93, opened 2023-09-08
Minor compiler warning: unused-but-set-variable


lczech:
Hi there,

with gcc-13.1.0, I'm getting 

```
htscodecs/htscodecs/rANS_static16_int.h:316:9: warning: variable 'tab_size' set but not used [-Wunused-but-set-variable]
  316 |     int tab_size = 0, i, j, z;
      |         ^~~~~~~~
```

from [here](https://github.com/samtools/htscodecs/blob/11b5007ffb68bea9f6c777874a215e4187ce659a/htscodecs/rANS_static16_int.h#L316C9-L316C17).

The `tab_size` is only used in an assertion, see [here](https://github.com/samtools/htscodecs/blob/11b5007ffb68bea9f6c777874a215e4187ce659a/htscodecs/rANS_static16_int.h#L416-L417), and hence never read in release mode.

If this behavior is intended, I'd suggest to simply add a `(void) tab_size;` somewhere to silence that warning.

Cheers
Lucas

PR #94, opened 2023-09-09
Add tab_size void cast to silence compiler warning

lczech:
PR for https://github.com/samtools/htscodecs/issues/93

The commit is based on [htscodecs@11b5007](https://github.com/samtools/htscodecs/tree/11b5007ffb68bea9f6c777874a215e4187ce659a), which is the commit that the htscodecs submodule in [htslib](https://github.com/samtools/htslib/tree/5acbc150b2c12ea77111d845c06179507df976ae) is currently set to. Might need rebasing, as the master branch here does not seem to have that unused variable issue.

### DIFF
--- a/htscodecs/rANS_static16_int.h
+++ b/htscodecs/rANS_static16_int.h
@@ -415,6 +415,7 @@ static inline int encode_freq1(uint8_t *in, uint32_t in_size, int Nway,
 
     tab_size = cp - out;
     assert(tab_size < 257*257*3);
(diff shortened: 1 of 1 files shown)


jkbonfield:
Thank you for the bug report and PR, however it's already been fixed in the latest commit cb3cde80389acd14d38a6fd713d6347d9b25ce6c by removing the asserts entirely.  They were added to spot coding errors during initial development rather than bad data (as just crashing is never a good way to handle that), and are long past the point of being useful.

PR #96, opened 2023-09-20
Downfall mitigation via simulated gathers

jkbonfield:
Also incorporates the old #92.

Intel Downfall / GDS security flaw caused the issue of an updated microcode to mitigate the problem.  Unfortunately the consequence of this is that SIMD gather instructions are now considerably slower, particularly AVX2.  So much so that it turns out saving the SIMD register to memory, using traditional scalar code to do the memory fetches one at a time and then loading those scalar fetches back into the SIMD register, is *considerably* faster than the AVX2 gather microcode instruction.  The mind boggles, but there we are.  As it turns out I was already exploring these sort of things for AMD Zen4 too as that has (had) a far slower gather instruction than Intel.

We could do on-the-fly measurements and work out how far the gather is, so we know whether to use a simulated one or a real one, but it's muddled somewhat as the scalar code consists of many instructions which can get reordered by the compiler, so how efficient it is depends on the context.  Similarly with a single instruction, the cost of it depends on how many other instructions we can do while waiting on the result - whether we can hide the instruction latency basically.

For now, I've disabled real gathers entirely, but they can be reenabled using `-DUSE_GATHER` cpp define.

The plot below shows the speed for an Intel CascadeLake CPU with the microcode patch applied, booted as-is and with the "mitigations=off" linux kernel parameter.  (That turns off spectre and meltdown too, but I've also benchmarked the simulated code with/without mitigations and demonstrated it's not impacted.)

image

Some comments:

0. All values are *rates* in MB/s, so higher is better.  Blue is encode, using the left Y axis.  Green is decode, using the right Y axis.
1. master-on and master-off are mitigations=on and mitigations=off with the master branch, using actual SIMD gathers.  SimG-on is simulated gathers, with mitigations=on (ie Downfall bug fixed).
2. We see for AVX2 encode speed is unaffected, but with AVX512 we do get an encode performance hit.  Both have decode performance hits, but AVX2 more so.
3. This PR (SimG-on) rescues most of the speed lost for AVX2.
4. This PR rescues *some* of the speed for AVX512, but it's quite challenging, particularly AVX512 order-1 encode.
5. If AVX512 is slower than AVX2, we don't have to use it.  We already had logic for that with AMD, so this has been extended.  Hence AVX512 slow encodes are mostly mitigated by using the AVX2 encoder.

With that final mitigation in place, the performance hit by using simulated gathers over actual gathers on an Intel machine without the new microcode (master=off) is in the 10-20% bracket.  That's probably acceptable, given the performance loss of not changing this code could be 65% (2.2x slower for AVX2 decode).

Also included below is the same system compiled using gcc 13.

image

I'll produce some AMD Zen4 (Genoa) plots soon.

### DIFF
--- a/htscodecs/rANS_static32x16pr_avx2.c
+++ b/htscodecs/rANS_static32x16pr_avx2.c
@@ -566,8 +566,6 @@ unsigned char *rans_uncompress_O0_32x16_avx2(unsigned char *in,
         // packs_epi16 ponmlkji ponmlkji  hgfedcba hgfedcba
         sv1 = _mm256_packus_epi32(sv1, sv2);
         sv1 = _mm256_permute4x64_epi64(sv1, 0xd8);
-        __m256i Vv1 = _mm256_cvtepu16_epi32(_mm_loadu_si128((__m128i *)sp));
-        sv1 = _mm256_packus_epi16(sv1, sv1);
 
         // Protect against running off the end of in buffer.
         // We copy it to a worst-case local buffer when near the end.
@@ -576,6 +574,9 @@ unsigned char *rans_uncompress_O0_32x16_avx2(unsigned char *in,
             sp = (uint16_t *)overflow;
             cp_end = overflow + sizeof(overflow) - 64;
         }
+
+        __m256i Vv1 = _mm256_cvtepu16_epi32(_mm_loadu_si128((__m128i *)sp));
+        sv1 = _mm256_packus_epi16(sv1, sv1);
 
         // c =  R[z] < RANS_BYTE_L;
 
@@ -596,14 +597,6 @@ unsigned char *rans_uncompress_O0_32x16_avx2(unsigned char *in,
         __m256i Yv1 = _mm256_slli_epi32(Rv1, 16);
         Vv1 = _mm256_permutevar8x32_epi32(Vv1, idx1);
         __m256i Yv2 = _mm256_slli_epi32(Rv2, 16);
-
-        // Protect against running off the end of in buffer.
-        // We copy it to a worst-case local buffer when near the end.
-        if ((uint8_t *)sp > cp_end) {
-            memmove(overflow, sp, cp_end+64 - (uint8_t *)sp);
-            sp = (uint16_t *)overflow;
-            cp_end = overflow + sizeof(overflow) - 64;
-        }
 
         // Shuffle the renorm values to correct lanes and incr sp pointer
         unsigned int imask2 = _mm256_movemask_ps((__m256)renorm_mask2);
--- a/htscodecs/rANS_static32x16pr_avx2.c
+++ b/htscodecs/rANS_static32x16pr_avx2.c
@@ -115,14 +115,41 @@ static inline __m256i _mm256_mulhi_epu32(__m256i a, __m256i b) {
 }
 #endif
 
-#if 0
-// Simulated gather.  This is sometimes faster as it can run on other ports.
+#ifndef USE_GATHER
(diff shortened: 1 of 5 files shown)


jkbonfield:
AMD Zen4 benchmarks:

image

image

Disabling mitigations basically has no impact here, so master-on and master-off doesn't matter.  What's important is the difference between master branch and this branch.

1. AVX2 order-0: no real change.
2. AVX2 order-1: slight gain in encode speed, slight loss in decode speed.  Nothing large.
3. AVX512 order-0: small encode speed lost (unused), and even smaller decode speed gain.  Note however the AVX2 encoder is faster and will be used anyway.
4. AVX512 order-1: small gain to both encode (but still unused) and decode speeds.

Nothing major basically, but it's good to see we haven't harmed non-Intel systems in this work.

htscodecs/rANS_static32x16pr_avx2.c
```diff
@@ -567,6 +589,14 @@ unsigned char *rans_uncompress_O0_32x16_avx2(unsigned char *in,
...
         __m256i Yv2 = _mm256_slli_epi32(Rv2, 16);
 
+        // Protect against running off the end of in buffer.
+        // We copy it to a worst-case local buffer when near the end.
+        if ((uint8_t *)sp > cp_end) {
+            memmove(overflow, sp, cp_end+64 - (uint8_t *)sp);
+            sp = (uint16_t *)overflow;
+            cp_end = overflow + sizeof(overflow) - 64;
+        }
+
```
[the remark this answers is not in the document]
jkbonfield: As suspected, if I remove the size limit check and bug fix the entropy.c code so it changes cpu properly, I can trigger this bug on master branch too with gcc, but not clang.

This chimes with the bug you found above, which was basically a consequence of removing the clang specific code path and making it work the same as gcc.  I'll try and figure out the correct bounds checks.  Thanks

htscodecs/rANS_static32x16pr_avx512.c
```diff
@@ -564,8 +640,45 @@ unsigned char *rans_compress_O1_32x16_avx512(unsigned char *in,
...
+            }
+            if (next_batch >= 0) {
+              __m128i c1_ = _mm_load_si128((__m128i *)&t32[next_batch][0]);
+              __m128i c2_ = _mm_load_si128((__m128i *)&t32[next_batch][16]);
+              c1 = _mm512_cvtepu8_epi32(c1_);
+              c2 = _mm512_cvtepu8_epi32(c2_);
+            }
+        }
+        if (next_batch < 0) {
+            c1 = _mm512_i32gather_epi32(iN1, in, 1);
```
daviesrob, line 675: You didn't make these `_mm512_i32gather_epi32x()`.  Was that deliberate?

jkbonfield: Yes.  Note the last argument is size, which is 1 here.  The _mm512_i32gather_epi32x as written is explicitly only doing 32-bit ints.  If I recall I did experiment with an alternative implementation doing size of 1 too, but it didn't help.

jkbonfield:
This has been hanging around long enough that it's being problematic to test the newer fuzzing fixes, so I have a branch which merges this, various bits from master and other outstanding PRs, plus a new commit to fix another fuzz testing bug:

https://github.com/jkbonfield/htscodecs/tree/sim3-tidy-fuzz2

The last commit there fixes the bug AVX512 gather mem bounds bug we discussed in the office, but it's not viable to do as a PR to master as it requires this PR first.  It's messy, but I'm hoping once all the other PRs get merged I can replace this PR with the new branch.

PR #97, opened 2023-09-21
Chandak overflow

jkbonfield:
Fixes a couple buffer overruns during encode in fqzcomp and name tokeniser.  Both codecs are primarily fuzzed with random input for decode, but encode matters a lot too.

The fqzcomp code failed when given an extreme case of 1 million 1bp reads.  The code incorrectly computed the worst case expansion by failing to account for (mainly) the 2-bits of average quality across the reads.  This has been fixed both in a revised buffer growth, but also by adding proper bounds checking to the range coder.

The name tokeniser failed when hitting 128 tokens due to a lack of checks in the encoder.  Could be triggered by a read name of 128 # symbols.

Thanks to @shubhamchandak94 for the bug report.

### DIFF
--- a/htscodecs/c_range_coder.h
+++ b/htscodecs/c_range_coder.h
@@ -31,13 +31,18 @@ typedef struct {
     uc *in_buf;
     uc *out_buf;
     uc *in_end;
+    uc *out_end;
+    int err;
 } RangeCoder;
 
 static inline void RC_SetInput(RangeCoder *rc, char *in, char *in_end) {
     rc->out_buf = rc->in_buf = (uc *)in;
     rc->in_end = (uc *)in_end;
 }
-static inline void RC_SetOutput(RangeCoder *rc, char *out) { rc->in_buf = rc->out_buf = (uc *)out; }
+
+// NB: call RC_SetOutput first, and then RC_SetOutputEnd
+static inline void RC_SetOutput(RangeCoder *rc, char *out) { rc->in_buf = rc->out_buf = (uc *)out; rc->out_end = NULL;}
+static inline void RC_SetOutputEnd(RangeCoder *rc, char *out_end) { rc->out_end = (uc *)out_end; }
 static inline char *RC_GetInput(RangeCoder *rc) { return (char *)rc->in_buf; }
 static inline char *RC_GetOutput(RangeCoder *rc) { return (char *)rc->out_buf; }
 static inline size_t RC_OutSize(RangeCoder *rc) { return rc->out_buf - rc->in_buf; }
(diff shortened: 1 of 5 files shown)


shubhamchandak94:
Thanks @jkbonfield. This fixes the specific segfault issues. There is one other issue I encountered which is not a buffer overflow issue. If the tok3 input is not in the human readable ASCII range (e.g., the input is `[1, 0]` where 0 is the NULL byte) then the encoding and decoding succeed but the decoded output replaces the 1 with 0. I wonder if checks can be added to prevent this or if a statement can be made about the valid inputs (I understand that the SAM standard imposes some constraints on read names). Thanks

jkbonfield:
Fixed.  It was the fast and loose way it detected end of name, given it's happy to have newline terminated (command line testing) or nul-terminated names (CRAM name block), and was treating it as two empty names instead.  I'm now more explicit in the validation on inputs.

Thanks

shubhamchandak94:
Thanks, that fixes it!

htscodecs/tokenise_name3.c
```diff
@@ -1464,11 +1468,17 @@ uint8_t *tok3_encode_names(char *blk, int len, int level, int use_arith,
 
     // Encode name
     for (i = j = 0; i < len; j=++i) {
-        while (i < len && blk[i] > '\n')
+        while (i < len && blk[i] >= ' ') // also catches top-bit set
```
daviesrob, line 1471: This won't work on platforms with unsigned char (e.g. ARM).

htscodecs/c_range_coder.h
```diff
@@ -61,13 +67,39 @@ static inline void RC_StartDecode(RangeCoder *rc)
...
     if (rc->in_buf+5 > rc->in_end) {
         rc->in_buf = rc->in_end; // prevent decode
         return;
     }
     DO(5) rc->code = (rc->code<<8) | *rc->in_buf++;
 }
 
+static inline void RC_ShiftLowCheck(RangeCoder *rc) {
+    if (rc->low < Thres || rc->Carry) {
+        if (rc->out_end && rc->out_buf + rc->FFNum >= rc->out_end) {
```
daviesrob, line 80: Strictly, to avoid UB this should be:
```suggestion
        if (rc->out_end && rc->FFNum >= rc->out_end - rc->out_buf) {
```

htscodecs/tokenise_name3.c
jkbonfield: Good spot.  I always forget that there are systems where char can be unsigned!  Fixed.

shubhamchandak94:
Thanks for looking into this and fixing! Will this be part of an upcoming release? Is there a timeline for that?

jkbonfield:
Given it's a security issue, I'd like to do a release soon.  However I also want the Intel Downfall mitigation patch merging first as it's a pretty big hit to performance after their microcode patch.  I don't have a time frame.  I'll try and get someone to review that PR.

PR #101, opened 2023-10-02
Remove divide by zero in rANS 4x8.

jkbonfield:
When given 0 bytes of input, we get a floating point exception in the code attempting to renormalise the frequencies.

The test harness can't trigger this case as 0 is EOF in the read loop, but that has been changed so a null file still calls the compression. Also tested the other entropy encoders, but they already coped fine.

Bug reported by Shubham Chandak

### DIFF
--- a/htscodecs/rANS_static.c
+++ b/htscodecs/rANS_static.c
@@ -96,7 +96,7 @@ unsigned char *rans_compress_O0(unsigned char *in, unsigned int in_size,
         free(out_buf);
         return NULL;
     }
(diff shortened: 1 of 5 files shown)


jkbonfield:
I fixed the UBSAN warning by adding a check in the library code, as it feels incorrect to be changing the behaviour of the test tool.

I'm not even convinced is it undefined behaviour, and suspect it's just a memory validity check that is too loose.  (I say this as the C99 standard has nothing to say on the matter as far as I could see.)  However it's a trivial fix and silencing warnings is a good thing for ease of fuzzing.

PR #103, opened 2023-10-04
A mixture of fuzz-testing improvements

jkbonfield:
- Fix rans_set_cpu to work a second time.  This only really matters for test/entropy.
- Add a test/entropy_fuzz.c  (@daviesrob you'll want this commit as it has the Makefile work, but you may want to edit the entropy_fuzz.c file itself.  Mine is just a copy of entropy.c, which is working and finding bugs)
- Fix the rans_set_cpu to once again support different CPU types for encoder and decoder.  We do test every codec, but we accidentally lost this ability to have half and half, which affected cross-validation between codecs.
- Small typo fix in entropy.c where it wasn't using RLE+PACK+SIMD.

### DIFF
--- a/tests/fqzcomp_qual_fuzzrt.c
+++ b/tests/fqzcomp_qual_fuzzrt.c
@@ -1,0 +1,137 @@
+/* Fuzz testing target. */
+/*
+ * Copyright (c) 2023 Genome Research Ltd.
+ * Author(s): James Bonfield
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice,
+ *       this list of conditions and the following disclaimer.
+ *
+ *    2. Redistributions in binary form must reproduce the above
+ *       copyright notice, this list of conditions and the following
+ *       disclaimer in the documentation and/or other materials provided
+ *       with the distribution.
+ *
+ *    3. Neither the names Genome Research Ltd and Wellcome Trust Sanger
+ *       Institute nor the names of its contributors may be used to endorse
+ *       or promote products derived from this software without specific
+ *       prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY GENOME RESEARCH LTD AND CONTRIBUTORS "AS
+ * IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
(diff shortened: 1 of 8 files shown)


tests/fqzcomp_qual_fuzzrt.c
```diff
@@ -0,0 +1,143 @@
...
+
+	// Semi random strat, but based on a few bits of input data
+	// for reproducability.
+	// This lets the fuzzer explore the parameter space itself.
+	int strat = in_size ? in[0] & 3 : 0;
+	char *comp = fqz_compress(3, s, (char *)in, in_size, &c_size,
+				  strat, NULL);
+	if (!comp) {
+	    fprintf(stderr, "REJECT FQZ %d to (null)n", (int)in_size);
+	    return 1;
```
daviesrob, line 119: If you want to reject this input, you should [return -1, not 1](https://www.llvm.org/docs/LibFuzzer.html#rejecting-unwanted-inputs).

tests/fqzcomp_qual_fuzzrt.c
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	char *uncomp = fqz_decompress(comp, c_size, &u_size, NULL, 0);
+	if (!uncomp) {
+	    free(comp);
+	    abort();
+	}
+	if (in_size != u_size) {
+	    free(comp);
```
daviesrob, line 128: You don't really need to free here (and it would aid debugging if you didn't).

tests/tokenise_name3_fuzzrt.c
```diff
@@ -0,0 +1,100 @@
...
+#include <time.h>
+
+#include "htscodecs/tokenise_name3.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *in, size_t in_sz) {
+    int level, arith;
+    char in2[8192];
+
+    // 4096 is default max size for libfuzzer anyway
+    if (in_sz > 8192) abort();
```
daviesrob, line 64: Return -1.

tests/fqzcomp_qual_fuzzrt.c
jkbonfield: I recall that conversation and that I meant to change it, but forgot. Thanks.

tests/fqzcomp_qual_fuzzrt.c
jkbonfield: It's to avoid leaking memory, but I guess it doesn't get to that stage of address sanitizer anyway so I can remove all the frees before abort here and elsewhere.

jkbonfield:
I've done the corrections and force pushed up again.

PR #107, opened 2023-10-09
Fix tok3 decode bounds checking.

jkbonfield:
We were one byte too short on our bounds checking, which meant a serialised data stream that ended with a duplicate token failed the "o+2 >= sz" check.

This could be reproduced with a pair of names "00x\n12\n".


daviesrob:
Note that the suggested input `"00x\n12\n"` only caused a problem if level was less than 3.

PR #112, opened 2023-10-19
Add explicit -mpopcnt to avx2 and avx512 builds.

### LINKED ISSUE #109: zig cc complains about inlining a function into anlother without support for 'popcnt'
When trying to use `CC='zig cc'` the build fails with problems such as:
```
/bin/bash ../libtool  --tag=CC   --mode=compile zig cc -DHAVE_CONFIG_H -I. -I../../htscodecs -I..    -mavx2 -Wall -g -O3 -MT librANS_static32x16pr_avx2_la-rANS_static32x16pr_avx2.lo -MD -MP -MF .deps/librANS_static32x16pr_avx2_la-rANS_static32x16pr_avx2.Tpo -c -o librANS_static32x16pr_avx2_la-rANS_static32x16pr_avx2.lo `test -f 'rANS_static32x16pr_avx2.c' || echo '../../htscodecs/'`rANS_static32x16pr_avx2.c
libtool: compile:  zig cc -DHAVE_CONFIG_H -I. -I../../htscodecs -I.. -mavx2 -Wall -g -O3 -MT librANS_static32x16pr_avx2_la-rANS_static32x16pr_avx2.lo -MD -MP -MF .deps/librANS_static32x16pr_avx2_la-rANS_static32x16pr_avx2.Tpo -c ../../htscodecs/rANS_static32x16pr_avx2.c  -fPIC -DPIC -o .libs/librANS_static32x16pr_avx2_la-rANS_static32x16pr_avx2.o
../../htscodecs/rANS_static32x16pr_avx2.c:353:18: error: always_inline function '_mm_popcnt_u32' requires target feature 'popcnt', but would be inlined into function 'rans_compress_O0_32x16_avx2' that is compiled without support for 'popcnt'
  353 |         ptr16 -= _mm_popcnt_u32(imask4);
      |                  ^
```

zig cc bug, or something not described quite right in htscodecs?

jkbonfield:
Zig adds native support for the CPU being built on.  So if we support popcnt then we don't need an explicit -mpopcnt.  However adding any -m options appears to disable that native support, so -mavx2 doesn't enable -mpopcnt, yet our auto-detection previously claimed it was necessary.  This means building on a CPU with popcnt but no avx2 will fail to build (with zig) when targetting the avx2 platform.

Therefore the only reliable way to check is the combinations actively in use.  So if we want to compile code using popcnt + avx2 then we have to test -mpopcnt -mavx2 together.

It's problematic and tricky to manage in autoconf.

Fixes #109

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -129,68 +129,33 @@ dnl AC_CHECK_LIB([bsc], [bsc_compress], [
 dnl 	LIBS="-lbsc $LIBS"
 dnl 	AC_DEFINE([HAVE_LIBBSC],1,[Define to 1 if you have the libbsc library.])])
 
-dnl Count parts needed to build rANS_static32x16pr_sse4.c
-sse4_prerequisites=""
-
-dnl Check if we can use our SSSE3 implementations of rANS 32x16 codec.
-HTS_CHECK_COMPILE_FLAGS_NEEDED([ssse3], [-mssse3], [AC_LANG_PROGRAM([[
+dnl Check if we can use our SSE4.1 too.
+dnl Our SSE4 codec uses SSE4.1, SSSE3 (shuffle) and POPCNT, so we check all 3
+dnl together.  This helps Zig builds which don't work well if we test each
+dnl individually.
+HTS_CHECK_COMPILE_FLAGS_NEEDED([sse4.1], [-msse4.1 -mssse3 -mpopcnt], [AC_LANG_PROGRAM([[
 	  #ifdef __x86_64__
 	  #include "x86intrin.h"
 	  #endif
 	]],[[
 	  #ifdef __x86_64__
 	  __m128i a = _mm_set_epi32(1, 2, 3, 4), b = _mm_set_epi32(4, 3, 2, 1);
-	  __m128i c = _mm_shuffle_epi8(a, b);
-	  return *((char *) &c);
-	  #endif
-	]])], [
-        MSSSE3="$flags_needed"
-	sse4_prerequisites="o$sse4_prerequisites"
-	AC_SUBST([MSSSE3])
-        AC_DEFINE([HAVE_SSSE3],1,[Defined to 1 if rANS source using SSSE3 can be compiled.])
-])
-
-dnl Check if we can use popcnt instructions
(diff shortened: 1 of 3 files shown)


daviesrob:
This doesn't look correct.  The small test programs need to be changed so that they use instructions that are gated by all the flags being tested for.  That way the test will only return an empty set if it really is true that none of the flags is necessary.  It will potentially return more flags than strictly necessary if it does require some of them, but that shouldn't be a problem.  The individual `ssse3` and `popcnt` tests could then be removed as they'd no longer be necessary.

The change to `m4/hts_check_compile_flags_needed.m4` also looks incorrect.  The cached value should be `none`, `unsupported` or the flags needed.  Among other things, changing this makes the output from `./configure` misleading:
```
checking C compiler flags needed for avx512f... avx512f
```

It does need a small fix to make the multiple flags work, but just adding quotes when setting `CACHEVAR` works:

```diff
--- a/m4/hts_check_compile_flags_needed.m4
+++ b/m4/hts_check_compile_flags_needed.m4
@@ -50,7 +50,7 @@ AC_CACHE_CHECK([_AC_LANG compiler flags needed for $1], CACHEVAR, [
     [ax_check_save_flags=$[]_AC_LANG_PREFIX[]FLAGS
      _AC_LANG_PREFIX[]FLAGS="$[]_AC_LANG_PREFIX[]FLAGS $6 $2"
      AC_LINK_IFELSE([m4_default([$3],[AC_LANG_PROGRAM()])],
-       [AS_VAR_SET(CACHEVAR,[$2])],
+       [AS_VAR_SET(CACHEVAR,["$2"])],
        [AS_VAR_SET(CACHEVAR,[unsupported])])
      _AC_LANG_PREFIX[]FLAGS=$ax_check_save_flags])])
 AS_VAR_IF(CACHEVAR,unsupported, [
```

Then you get:
```
checking C compiler flags needed for avx512f... -mavx512f -mpopcnt
```

daviesrob:
The latest version of this PR seems to work OK, although I found that with the latest zig (0.11.0) `./configure CC="zig cc"` decided no flags were needed and worked fine, even with the original tests.  The changes here do rescue you if you set architecture-related CFLAGS though, e.g. `./configure CC="zig cc" CFLAGS="-g -O3 -mavx512f"`.

PR #116, opened 2024-04-22
Add cpuid checks behind an HAVE_CPUID check.

### LINKED ISSUE #115: Undefined symbols ___cpuid_count ___get_cpuid_max building on Mac OS X 10.7
htslib 1.20 fails to build on Mac OS X 10.7.5 with the version of Apple's fork of the clang compiler (`Apple LLVM version 4.2 (clang-425.0.28) (based on LLVM 3.2svn)`) that comes with Xcode 4.6.3 (the last version compatible with Mac OS X 10.7) due to:

```
Undefined symbols for architecture x86_64:
  "___cpuid_count", referenced from:
      _htscodecs_tls_cpu_init in rANS_static4x16pr.o
  "___get_cpuid_max", referenced from:
      _htscodecs_tls_cpu_init in rANS_static4x16pr.o
ld: symbol(s) not found for architecture x86_64
```

As far as I know, versions of Apple clang less than 503 have this deficiency. Using a newer version of clang is a workaround—either a newer version of Apple clang on newer versions of macOS, or a newer version of llvm.org clang installed manually.

jkbonfield:
Fixes #115.

Testing compiler versions is tricky, especially given some vendors wrap up standard compilers as their own, but change version numbers (Apple, Intel) and compilers also often masquerade as another (eg setting GNUC).

Instead we just check if it links.  Note there's no guard around this here for x86-64, but that's done in the code itself.  (In theory we may have cpuid on another architecture that we wish to use at some point.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -120,6 +120,25 @@ AS_IF([test "$ax_cv_builtin_prefetch" = "yes"],
            [Define to 1 if you have a __builtin_prefetch])],
 	[])
 
+dnl Check is cpuid works, needed by rANS_static4x16pr.c
+AC_MSG_CHECKING(for cpuid support)
+AC_LINK_IFELSE(
+    [AC_LANG_PROGRAM([
+        #include <cpuid.h>
+        #include <stddef.h>
+    ], [
+        unsigned int eax = 0, ebx = 0, ecx = 0, edx = 0;
+        int level = __get_cpuid_max(0, NULL);
+        if (level >= 1)
+            __cpuid_count(1, 0, eax, ebx, ecx, edx);
+    ])], [
+        AC_MSG_RESULT(yes)
+        AC_DEFINE(HAVE_CPUID, 1, [Set to 1 to use cpuid])
+    ], [
+        AC_MSG_RESULT(no)
+    ]
+)
+
 dnl AC_CHECK_LIB([lzma], [lzma_easy_buffer_encode], [
 dnl 	LIBS="-llzma $LIBS"
 dnl 	AC_DEFINE([HAVE_LIBLZMA],1,[Define to 1 if you have the liblzma library.])])
--- a/configure.ac
+++ b/configure.ac
@@ -120,24 +120,8 @@ AS_IF([test "$ax_cv_builtin_prefetch" = "yes"],
            [Define to 1 if you have a __builtin_prefetch])],
 	[])
 
-dnl Check is cpuid works, needed by rANS_static4x16pr.c
-AC_MSG_CHECKING(for cpuid support)
-AC_LINK_IFELSE(
-    [AC_LANG_PROGRAM([
-        #include <cpuid.h>
-        #include <stddef.h>
-    ], [
-        unsigned int eax = 0, ebx = 0, ecx = 0, edx = 0;
-        int level = __get_cpuid_max(0, NULL);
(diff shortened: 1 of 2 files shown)


jkbonfield:
Problems still to consider.

How does this work from htslib, which although having this as a sub-package doesn't run the configure script ind instead builds the source direct.

1. Htslib's ./configure - it can just copy this logic and add `HAVE_CPUID` into its own config.h
2. Htslib's Makefile (configureless) could just copy the existing ifdef magic to the config.h stub it creates, so it replicates current behaviour (with the same limitations on portability).  Or we just define it to be true and accept if it fails then don't do that - run configure!

Htslib builds from source have htscodecs as a submodule, so both can update in sync.
Htslib can also link against an external htscodecs library, but the configure issue is then solved there already, so it's not a problem.

jmarshall:
I don't have an x86 machine handy to check on, but could this just be done with the built-in generic tests? e.g.

```
AC_CHECK_DECLS([__get_cpuid_max, __cpuid_count], [], [], [[#include <cpuid.h>]])
```

jkbonfield:
Good thought.  Possibly.  It'd need validating whether it also solves the problem on the Mac, but given it was a compiler warning about missing prototype as well as a link error it seems likely.  I'll try it out.

jkbonfield:
So... pros and cons.  The configure.ac is simpler: just 1 line as you suggest.  However the code that uses it is now ugly, made more so by the 0/1 nature rather than defined/not-defined as used elsewhere in autoconf.

```
#if defined(__x86_64__) && \
    defined(HAVE_DECL___CPUID_COUNT)   && HAVE_DECL___CPUID_COUNT && \
    defined(HAVE_DECL___GET_CPUID_MAX) && HAVE_DECL___GET_CPUID_MAX
...
```

Plus these will need to be added to the htslib Makefile too for when not using configure.

Thoughts?

jkbonfield:
I added those changes as a separate commit anyway, so we can either squash them together which means we get the new mechanism, or drop that commit and keep my previous method.  Both appear to work fine under linux.  I'll validate MinGW

jmarshall:
> Thoughts?

If you fill in the actions-if-found and/or actions-if-not-found arguments, you can have your own simpler choice of macros to be defined instead of getting the default actions.

jkbonfield:
What's the plan for htslib?

We can put in a similar autoconf rule easily enough as it's minimal code to add.  There's the Makefile though for when configure isn't used.  If we do nothing it should continue to build, just defaulting without optimised implementations.  I'm not sure I like that (but frankly I've never liked our ability to build without first running configure either as it's just a needless pitfall).  Instead I would advise hard coding the macros to be "on", so the default non-configured is just to assume this stuff works.  That's a reasonable assumption which should hold true everywhere given it still has an x86-64 CPU check anyway.

daviesrob:
I expect htslib `make` will probably end up setting these on `x86_64`.  It makes quite a few assumptions already, so I think this would be reasonable given that we can let `configure` override it if necessary.

jkbonfield:
My point was we don't need to have x86_64 checks in the Makefile as they're already in the source files.  We can just define these to 1 on all platforms and it'll have no consequences elsewhere.  We don't need extra fragile or complex code for platform detection when not using autoconf.

PR #119, opened 2024-05-14
Avoid disabled typedef in FreeBSD-specific code

jmarshall:
Since 6dadc2a8d42d4d101b0bc96807147fce96fe0ddf / samtools/htslib#1628, this code is compiled with `_XOPEN_SOURCE` set, usually to 600.

Setting this causes FreeBSD's `<sys/types.h>` not to define these old BSD-style typenames such as `u_long`. Happily we can just write it out as the corresponding native type — fortunately [`u_long`'s definition](https://cgit.freebsd.org/src/blame/sys/sys/types.h#n48) has not changed in 30+ years…

### DIFF
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -1043,11 +1043,11 @@ static inline int have_neon(void) {
 #elif defined(__APPLE__)
     return 1;
 #elif defined(__FreeBSD__) && defined(__arm__)
-    u_long cap;
+    unsigned long cap;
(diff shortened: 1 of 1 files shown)


jmarshall:
That is, when compiled via HTSlib's build infrastructure, _rANS_static4x16pr.c_ gets compiled with `_XOPEN_SOURCE` and needs this patch. I haven't tested whether `_XOPEN_SOURCE` is set when htscodecs is compiled standalone using its own build infrastructure — but in any case, the patch won't hurt.

daviesrob:
Thanks, the change looks fine.

PR #121, opened 2024-06-24
Add windows build via github Actions

jkbonfield:
Also fix freebsd to be up to date as it broke CI.


jkbonfield:
Replaced by #123

PR #122, opened 2024-06-25
sample PR

jkbonfield:
test of PR against a branch in samtools/htscodecs that already has a functioning GitHub Actions config.

### DIFF
--- a/.github/workflows/unix-build.yml
+++ b/.github/workflows/unix-build.yml
@@ -1,0 +1,48 @@
+# Various unix builds
+name: unix CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, ubuntu-20, macos-latest]
+    
+    defaults:
+      run:
+        shell: bash {0}
+
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
+    - name: macOS specific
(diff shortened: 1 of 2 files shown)


.github/workflows/unix-build.yml
```diff
@@ -0,0 +1,36 @@
...
+      run:
+        shell: bash {0}
+
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
+    - name: Mac install dependencies
+      if: ${{ matrix.os == 'macos-latest' }}
```
jmarshall, line 23: I'd suggest [`runner.os`](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context) instead. Its value is fixed, whereas next month you might want the matrix entry to be `macos-14` or so.

```suggestion
      if: runner.os == 'macOS'
```

(IIRC `==` is case insensitive anyway, but might as well use the documented capitalisation.)

jkbonfield: Thanks.  At the moment this is very exploratory.  Partly an attempt to avoid AppVeyor as it's getting slower and slower to launch Windows MinGW builds, and partly to offload some of the workload with give to cirrus-ci as it has a maximum amount of CPU we can use, has a reduced number of concurrent jobs, and we're low on priority in the free tier.  (Or we get our bill payers to consider such tooling as something to pay for instead of just the usual corporate-ware.)

Current wrestle: submodules and gitattributes.  Shrug.  Should work!  Will beat it around the head some more :)

PR #123, opened 2024-06-25
Github Actions configuration

jkbonfield:
NB this is a PR from samtools/htscodecs itself as a test to see if it then permits the new CI to run as validation.  (If not, you can see the results of the CI here anyway.)

- Remove AppVeyor (config has gone and temporarily deactivated the execution of it in the repo settings)
- Added GitHub Actions workflow for MinGW
- Added workflow for linux / macos basic builds, therefore reducing the number of concurrent cirrus-ci jobs needed.


.github/workflows/unix-build.yml
```diff
@@ -0,0 +1,45 @@
...
+        autoreconf -i
+        ./configure CC="gcc -fsanitize=address,undefined"
+
+    - name: Compile
+      run: |
+        make -j6
+
+    - name: Check
+      run: |
+        make check || cat tests/test-suite.log
```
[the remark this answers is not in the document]
daviesrob:
```suggestion
        make check || ( cat tests/test-suite.log ; false )
```

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,37 @@
...
+        export MSYSTEM=MINGW64
+        autoreconf -i
+        ./configure --disable-shared CFLAGS='-g -O3 -D_XOPEN_SOURCE=600  -Wno-char-subscripts'
+        make -j32
+    - name: Check htscodecs
+      shell: msys2 {0}
+      run: |
+        export PATH=/mingw64/bin:$PATH
+        export MSYSTEM=MINGW64
+        make check
```
daviesrob, line 36:
```suggestion
        make check || ( cat tests/test-suite.log ; false )
```

daviesrob (commented):
I've suggested changes to the `Check` sections that write out `tests/test-suite.log` on failure while ensuring the failed state is reported correctly.  I would have pushed these myself, but I'm running into some GitHub permissions problems around changing workflows...

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,37 @@
...
+          mingw-w64-x86_64-toolchain
+          mingw-w64-x86_64-autotools
+          mingw-w64-x86_64-tools-git
+    - name: Compile htscodecs
+      shell: msys2 {0}
+      run: |
+        export PATH=/mingw64/bin:$PATH
+        export MSYSTEM=MINGW64
+        autoreconf -i
+        ./configure --disable-shared CFLAGS='-g -O3 -D_XOPEN_SOURCE=600  -Wno-char-subscripts'
```
daviesrob, line 29: Why is `-Wno-char-subscripts` here?  It's not the sort of error that should be ignored.  Also, all builds should use `-Wall -Werror` as we know from experience that warnings get ignored unless they cause the whole test to fail.

jkbonfield: It was added because there were a large number of errors coming from the original (non mingw-w64) compiler that the github action was using on Windows.  None of the unix ones reported these, so for now I disabled it so I could break the problem down into one thing at a time - in this case getting windows working.

However I suspect they've now gone again as I realised my other errors came down to PATH and finding the wrong compiler.

I think the errors came from differing implementations of ctype, with the other compiler using an array based lookup.  Eg see https://github.com/jkbonfield/htscodecs/actions/runs/9644267159/job/26596006462

It's probably a genuine thing, but best solved via another PR.

jkbonfield: I've confirmed that `-Wno-char-subscripts` is no longer needed anyway even with the ctype problems unpatched.   However I fixed the ctype checking in #124.  Note I found the same problems in htslib (once), samtools(~10) and bcftools (100ish).  This is no doubt why it's worked around on linux, and I suspect several on many other platforms, as it's just an endemic problem and frankly a sign of a terrible clash of API design in C ctype vs str* functions.

I enabled -Wall -Werror.  (These were absent simply because we missed them off before in appveyor and I was copying that config.)  I could try using `configure --enable-werror` but I'm not sure when that appeared (we don't use it in htslib, but it's perhaps an automake thing) and apparently unlike htslib we don't explode if we attempt to define `-Werror` during configure anyway, presumably due to the limited range of tests we use.  We can add it later if it's a problem for e.g. cross-compiling.

PR #124, opened 2024-07-02
Use unsigned chars for ctype macros.

jkbonfield:
On linux these tables are 384 wide, to accommodate -128 to +255. However this is an implementation defined security improvement and the specification is such that the parameter to e.g. isalpha should be unsigned.

Detected with the stock windows gcc install (mingw-w64 didn't complain).

See https://github.com/samtools/htscodecs/pull/123#discussion_r1662357349 for background.

### DIFF
--- a/htscodecs/tokenise_name3.c
+++ b/htscodecs/tokenise_name3.c
@@ -765,19 +765,20 @@ static int encode_name(name_context *ctx, char *name, int len, int mode) {
         }
 
         /* Determine data type of this segment */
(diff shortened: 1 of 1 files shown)


jkbonfield:
Note Freebsd is failing because the cirrus-ci runner changed from 13.2 to 14.0.  It's fixed in #123 so I don't see the need for causing conflicts by fixing it elsewhere too.  If we merge this second then the rebased version should pass.

PR #127, opened 2024-09-02
Improve checking of *out_size in rans4x16 & arith

jkbonfield:
These should be set to a minimum of at least rans_compress_bound_4x16() or arith_compress_bound()  but as they're user supplied parameters we ought to consider the possibility of them being smaller and the potential for an encode failure.  (Although arguably it's a user error too as we're not using the API as intended.)

This happens in many places, but not consistently when very small sizes are given such that we don't have room for the compression header meta-data.

### DIFF
--- a/htscodecs/arith_dynamic.c
+++ b/htscodecs/arith_dynamic.c
@@ -733,15 +733,17 @@ unsigned char *arith_compress_to(unsigned char *in,  unsigned int in_size,
     unsigned int c_meta_len;
     uint8_t *rle = NULL, *packed = NULL;
 
-    if (in_size > INT_MAX) {
+    if (in_size > INT_MAX || (out && *out_size == 0)) {
         *out_size = 0;
         return NULL;
     }
 
     if (!out) {
         *out_size = arith_compress_bound(in_size, order);
-        if (!(out = malloc(*out_size)))
+        if (!(out = malloc(*out_size))) {
+            *out_size = 0;
             return NULL;
+        }
     }
     unsigned char *out_end = out + *out_size;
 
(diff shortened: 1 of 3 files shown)


jkbonfield:
Another round of updates squashed in.  I expect there may be issues with other codecs too, but we don't have to solve every problem in every PR.  This focuses only on the CRAM 3.1 rans and arith coders.

I have a (not committed as it's hacky) little test which constructs (ABAC)* repetition and then permits me to compress using various "order" options, with manually specified ranges of out_size buffers.  Note this is NOT how the API is meant to work.  You should call the appropriate function (eg `arith_compress_bound`) with an input buffer size and the order byte and it'll give you the worst case size of memory needed to compress the data, and it's up to you to ensure you provide a buffer that big.

We do just that in all our use cases, so this is not a bug anywhere for htslib, samtools, etc.  However if you're a trouble maker and allocate a buffer too small then we now bail out correctly.  Obviously you can still allocate it too small and say it's larger in which case it dies, but that's no different than lying to memcpy etc.   Both are API usage errors, but the former at least should have some chance of spotting memory errors before they happen as we've told it how big our (undersized) buffer was.

jkbonfield:
Somewhat unrelated, I also changed it to always return out_size=0 when failing, as well as returning NULL.  I don't know if that's useful or not, but it seemed tidier.  It's neither here nor there I guess.

htscodecs/arith_dynamic.c
```diff
@@ -843,20 +862,27 @@ unsigned char *arith_compress_to(unsigned char *in,  unsigned int in_size,
...
+
                 olen2 = *out_size - (out2 - out);
                 //fprintf(stderr, "order=%d m=%d\n", order&3, m[MIN(i,4)][j]);
                 if ((order&3) == 0 && (m[MIN(i,3)][j]&1))
                     continue;
 
                 arith_compress_to(transposed+idx[i], part_len[i],
                                   out2, &olen2, m[MIN(i,3)][j] | X_NOSZ);
-                if (best_sz > olen2) {
+                if (olen2 && best_sz > olen2) {
```
daviesrob, line 875: I'm not sure why you didn't check for a NULL return value here, as it would have needed fewer changes.  Anyway, checking `olen2` should now work here. However the `arith_compress_to()` recursion at line 888 assumes that it will work the second time round - I'm not sure if you could completely guarantee that as it could try to allocate memory for some of the options.  For safety it should probably be checked to ensure it worked.

htscodecs/rANS_static4x16pr.c
```diff
@@ -1199,11 +1201,20 @@ unsigned char *rans_compress_to_4x16(unsigned char *in, unsigned int in_size,
         int N = (order>>8) & 0xff;
         if (N == 0) N = 4; // default for compatibility with old tests
 
+        if (N > in_size)
+            N = in_size;
+
+        if (N > 255) {
```
daviesrob, line 1207: If this happens your CPU is broken, as `N` is masked above.  Note that this code isn't the same as the version in `arith_compress_to()`, where `N` currently could be greater than 255.

htscodecs/arith_dynamic.c
jkbonfield: The reason for it is traditionally it's been acceptable for some compression methods to fail, eg attempting to stripe with an indivisible number of components.  (This may have been resolved now, but I'd have to check.)  The philosophy though is don't error if any call fails, but error if all fails.  It doesn't matter much how we determine failure (return value vs length), but I wanted to make length return 0 on failure anyway so they become the same thing really.  Perhaps a NULL check is easier to understand though (and safer incase of paths I missed or may miss in the future) so I'll consider changing it.

htscodecs/rANS_static4x16pr.c
jkbonfield: And indeed the specification states it's one byte!  This is just an accident of duplication between the rans vs arith, where one is mistakenly not doing `& 0xff`.  Will fix.  Thanks

htscodecs/arith_dynamic.c
jkbonfield: Resolved with belt and braces both NULL and olen2.  It's not a critical part of the algorithm speed and it protects against future mistakes.

PR #134, opened 2025-05-13
UUID4 name tokeniser plus improvements to mixed data sets.

jkbonfield:
1.  UUID4 improvements.

With this change on my local corpus:

```
htscodecs-corpus/names/01.names     169625
htscodecs-corpus/names/02.names     417607
htscodecs-corpus/names/03.names     49868
htscodecs-corpus/names/05.names     288460
htscodecs-corpus/names/08.names     42594
htscodecs-corpus/names/09.names     255867
htscodecs-corpus/names/10.names     307270
htscodecs-corpus/names/20.names     139707
htscodecs-corpus/names/_139.names   86966   *
htscodecs-corpus/names/_shuf.names  638699  *
htscodecs-corpus/names/nv.names     329535
htscodecs-corpus/names/nv2.names    87802
htscodecs-corpus/names/rr.names     91103
htscodecs-corpus/names/uuid4+.names 153556  *
htscodecs-corpus/names/uuid4.names  153439  *
```

Master:

```
htscodecs-corpus/names/01.names     169625
htscodecs-corpus/names/02.names     417607
htscodecs-corpus/names/03.names     49981
htscodecs-corpus/names/05.names     288460
htscodecs-corpus/names/08.names     42594
htscodecs-corpus/names/09.names     255867
htscodecs-corpus/names/10.names     307270
htscodecs-corpus/names/20.names     139707
htscodecs-corpus/names/_139.names   100798  *
htscodecs-corpus/names/_shuf.names  710366  *
htscodecs-corpus/names/nv.names     329535
htscodecs-corpus/names/nv2.names    87802
htscodecs-corpus/names/rr.names     91103
htscodecs-corpus/names/uuid4+.names 187560  *
htscodecs-corpus/names/uuid4.names  242453  *
```

For uuid4 see #132.  The difference between uuid4 and uuid4+ is the latter has "-flowcell-10" appended to every line.  The old code had explicit check for ONT UUIDs, but not bare ones without an extension, so it needed 37 chars and not 36.  Hence master compresses the data with "-flowcell-10" better than without.  Regardless of that however, both compress better still by using an array of CHAR tokens instead of separate string-char-string-char-... combinations (due to nuls).

2. The other change can be seen in _139.names (a mix of 5k names from 01, 03 and 09.names) and _shuf (a mix of all of them).  This still isn't optimal, but it's considerably closer.

We can demonstrate this easily with two names of foo:<num> and bar:<num> where <num> is +1 in foo and +1-16 in bar.  

```
perl -e '$b1=1;$b2=1;for ($i=0;$i<5000;$i++) {print "foo:$b1\n";print "bar:$b2\n";$b1++;$b2+=1+int(rand(16))}' > _
```

Theoretically foo should have no information per record other than the initial meta-data describing the format.
Bar however has 4 bits of ddelta offset per record, so has 2.5Kb of information to encode.  Mixing them should not increase the size as we are strictly alternating and the diff distance is a fixed -2.

We can do `cat _ | tokenise_name` vs `egrep foo _ | ...` and `egrep bar _ | ...` to validate the mixed vs separate names.
Master gives 13300, 86 and 2603 respectively.
This PR gives 2684, 86 and 2603 repsectively.

The change achieves this by improving the returned node rather than -1, which defaults to the previous record.  It was previously always delta bar against foo and foo against bar.

### DIFF
--- a/htscodecs/tokenise_name3.c
+++ b/htscodecs/tokenise_name3.c
@@ -610,12 +610,16 @@ int search_trie(name_context *ctx, char *data, size_t len, int n, int *exact, in
         prefix_len = 6;  // IonTorrent
         *fixed_len = 6;
         *is_fixed = 1;
-    } else if (l > 37 && d[f+8] == '-' && d[f+13] == '-' && d[f+18] == '-' && d[f+23] == '-' &&
-               ((d[f+0] >= '0' && d[f+0] <='9') || (d[f+0] >= 'a' && d[f+0] <= 'f')) &&
-               ((d[f+35] >= '0' && d[f+35] <='9') || (d[f+35] >= 'a' && d[f+35] <= 'f'))) {
+    } else if (l >= 36
+               && d[f+8]=='-' && d[f+13]=='-' && d[f+18]=='-' && d[f+23]=='-'
+               && isxdigit((uint8_t)d[f+0])  && isxdigit((uint8_t)d[f+7])
+               && isxdigit((uint8_t)d[f+9])  && isxdigit((uint8_t)d[f+12])
+               && isxdigit((uint8_t)d[f+14]) && isxdigit((uint8_t)d[f+17])
+               && isxdigit((uint8_t)d[f+19]) && isxdigit((uint8_t)d[f+22])
+               && isxdigit((uint8_t)d[f+24]) && isxdigit((uint8_t)d[f+35])) {
         // ONT: f33d30d5-6eb8-4115-8f46-154c2620a5da_Basecall_1D_template...
-        prefix_len = 37;
-        *fixed_len = 37;
+        prefix_len = 36;
+        *fixed_len = 36;
         *is_fixed = 1;
     } else {
         // Check Illumina and trim back to lane:tile:x:y.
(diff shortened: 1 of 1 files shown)


htscodecs/tokenise_name3.c
```diff
@@ -610,12 +610,16 @@ int search_trie(name_context *ctx, char *data, size_t len, int n, int *exact, in
         prefix_len = 6;  // IonTorrent
         *fixed_len = 6;
         *is_fixed = 1;
-    } else if (l > 37 && d[f+8] == '-' && d[f+13] == '-' && d[f+18] == '-' && d[f+23] == '-' &&
-               ((d[f+0] >= '0' && d[f+0] <='9') || (d[f+0] >= 'a' && d[f+0] <= 'f')) &&
-               ((d[f+35] >= '0' && d[f+35] <='9') || (d[f+35] >= 'a' && d[f+35] <= 'f'))) {
+    } else if (l >= 36
+               && d[f+8]=='-' && d[f+13]=='-' && d[f+18]=='-' && d[f+23]=='-'
```
rhpvorderman, line 614: `d[f+14]` is always `4` in uuid4. See [wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)). So this can also be added as a check.

daviesrob: We tried some alternate UUID versions.  This change seems to be beneficial even with UUIDs with a common prefix (like version 7), so it seems best to leave it accepting all UUID versions.

PR #137, opened 2025-06-24
Protection from attempting to do SIMD rans encoding on small data

jkbonfield:
We can't just change do_simd as we may then have a combination of SIMD (pack or rle-meta) and non-SIMD (rle-meta or rle-data).  So instead we remove the ability to do SIMD and call again.

This only happens when we have large blocks of data that enable SIMD mode which then turn out to be heavily compressible after PACK or RLE to yield a small sub-stream which isn't SIMD compressible.


htscodecs/rANS_static4x16pr.c
```diff
@@ -1385,6 +1389,14 @@ unsigned char *rans_compress_to_4x16(unsigned char *in, unsigned int in_size,
             int sz = var_put_u32(out+c_meta_len, out_end, in_size);
             c_meta_len += sz;
             *out_size -= sz;
+
+            if (do_simd && in_size < 32) {
+                free(packed);
+                return rans_compress_to_4x16(orig_in, orig_in_size,
+                                             out, out_size,
+                                             orig_order & ~(RANS_ORDER_X32 |
+                                                            RANS_ORDER_SIMD_AUTO));
```
daviesrob, line 1398: This works, but wouldn't it be simpler to just drop X32 mode and carry on:
```c
                do_simd = 0;
                out[0] &= ~RANS_ORDER_X32;
```
so that you avoid having to re-do all the RLE / packing work?  (And the same below...)

daviesrob:
Fixed a minor white-space change...

PR #140, opened 2025-09-15
Add cpuid checks for XSAVE, OSXSAVE and AVX

daviesrob:
AVX2 and AVX512F depend on AVX, XSAVE and OSXSAVE being present. If they are not, AVX2/AVX512F instructions may be blocked even though `cpuid()` reports them to be available.  Spotted when the AVX512 rANS codec crashed with illegal instructions on a virtual host that claimed to have AVX512F, but did not have OSXSAVE.

For comparison, see similar detection code in [libdeflate](https://github.com/ebiggers/libdeflate/blob/d65d5dbcc980236775d77ab6ea202c59852b7b7b/lib/x86/cpu_features.c#L181) and [gcc](https://gcc.gnu.org/git?p=gcc.git;a=blob;f=gcc/common/config/i386/cpuinfo.h;h=c93ea07239ad9f7b3c68ee3226a72582773abce7;hb=HEAD#l734).

### DIFF
--- a/htscodecs/rANS_static4x16pr.c
+++ b/htscodecs/rANS_static4x16pr.c
@@ -844,6 +844,33 @@ static int rans_cpu = 0xFFFF; // all
 #  define UNUSED
 #endif
 
+#if defined(__APPLE__)
+/*
+   MacOS before 12.2 (a.k.a. Darwin 21.3) had a bug that could cause random
+   failures on some AVX512 operations due to opmask registers not being restored
+   correctly following interrupts.  For simplicity, check that the major version
+   is 22 or higher before using AVX512.
+   See https://community.intel.com/t5/Software-Tuning-Performance/MacOS-Darwin-kernel-bug-clobbers-AVX-512-opmask-register-state/m-p/1327259
+*/
+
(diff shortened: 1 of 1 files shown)


jkbonfield:
Note this implementation is not the same as the ones you linked, which use XGETBV (in machine code as binutils may not support it) and then check XCR[0] for 0x06 (AVX2) or 0xE6 (AVX512).  The comments in libdeflate are more useful, but it has hard coded numeric values (6/e6), while gcc's implementation has no comments about why it's dropped to raw machine code but does have sensible macros explaining the magic numbers.

Now whether or not we need to care about these I don't know, but my instinct here is if we're fixing this to support the OSXSAVE flags then we ought to then actually query them rather than being binary on/off to see which flags have been filtered out by the OS.

daviesrob:
Added `XGETBV` lookup and a commit to disable AVX512 on versions of MacOS which had a bug that could cause some AVX512 operations to randomly fail.

The `XGETBV` commit can be tested by passing boot option `clearcpuid=avx` or `clearcpuid=avx512f` to the Linux kernel.  The MacOS one will need an Intel Mac.

PR #145, opened 2026-01-14
Prevent use of unaligned uint16_t pointers

daviesrob:
UBSan on clang complains about dereferencing unaligned `uint16_t`pointers even when the dereference is being done by `memmove()`. Strictly it's correct as even setting the pointer to such a value is undefined behaviour according to the standard.

Silence the noise by changing the pointers to `uint8_t` and adjusting the arithmetic on them as necessary.

Some bounds checks to ensure fast code won't read beyond the end of its input are also adjusted to prevent any possibility of generating an address beyond the limits of the input memory.

(`configure CFLAGS='-g -O2 -fsanitize=address,undefined -fno-sanitize-recover=all' LDFLAGS='-g -O2 -fsanitize=address,undefined'` to highlight the problem this solves.)


daviesrob:
I'm happy with your changes.

PR #150, opened 2026-07-06
Update deprecated names from the README

rhpvorderman:
When I looked at implementation details `encode_names` and `decode_names` where calling their `tok3_` named counterparts and there was a deprecation note in the comments.


jkbonfield:
Yes we prefer the tok3 prefix on those and I should have updated the (rather minimal) documentation accordingly.

Thanks.

